PR #1, opened 2022-10-19
better output format

ermalkaleci:


xlc:
This is temporary testing code and will be removed soon

PR #8, opened 2022-10-26
make wasm-executor default one

ermalkaleci:


src/task.ts
```diff
@@ -64,8 +62,9 @@ export class TaskManager {
   }
 
   runTask(taskId: number): Promise<void> {
-    if (this.#executorCmd && this.#executorCmd.length > 0) {
-      const cmd = `${this.#executorCmd} --runner-url=ws://localhost:${this.#listeningPort} --task-id=${taskId}`
+    const executorCmd = process.env.EXECUTOR_CMD || ''
```
xlc, line 65: better to pass the value instead of depends on a global variable
all access of process.env should be in the entry point for discovery purpose

package.json
```diff
@@ -11,9 +11,8 @@
 		"prepare": "husky install",
 		"start": "ts-node src/index.ts",
 		"build-wasm": "cd executor && wasm-pack --log-level=info build --target nodejs --no-default-features",
-		"build": "cd executor && cargo build --locked",
-		"test": "vitest",
-		"test:wasm": "WASM_EXECUTOR=true vitest",
+		"check": "cd executor && cargo check --locked",
+		"test": "vitest --silent",
```
xlc, line 15: add `test:native` and `dev:native` that uses `cargo run` for the runner

xlc (approved):

PR #32, opened 2022-10-31
option to pre-define or override state

ermalkaleci:


src/utils/set-storage.ts
```diff
@@ -0,0 +1,79 @@
...
+interface StorageKeyMaker {
+  meta: StorageEntryMetadataLatest
+  makeKey: (...keys: any[]) => StorageKey
+}
+
+const storageKeyMaker =
+  (registry: Registry, metadata: Metadata) =>
+  (section: string, method: string): StorageKeyMaker => {
+    const pallet = metadata.asLatest.pallets.filter((x) => x.name.toString() === section)[0]
+    assert(pallet)
```
xlc, line 19: should throw error with some description to indicate bad parameter

PR #48, opened 2022-11-11
refactor runtime version & improvements

ermalkaleci:


src/blockchain/block.ts
```diff
@@ -145,50 +142,49 @@ export class Block {
...
+      ]).then(([data, properties, chain, version]) => {
         const registry = new TypeRegistry(this.hash)
-        registry.setMetadata(new Metadata(registry, data))
         registry.setChainProperties(registry.createType('ChainProperties', properties))
+        registry.register(getSpecTypes(registry, chain, version.specName, version.specVersion))
+        registry.setHasher(getSpecHasher(registry, chain, version.specName))
+        registry.setMetadata(
+          new Metadata(registry, data),
+          undefined,
+          objectSpread<ExtDef>({}, getSpecExtensions(registry, chain, version.specName), {})
```
xlc, line 158: I am still not 100% sure why use `objectSpread` instead of the actual spread syntax

xlc (approved):

src/blockchain/block.ts
ermalkaleci: Not need. Replaced the rest with empty object and forgot to remove it

PR #80, opened 2022-12-10
export metadata

dzmitry-lahoda:


obsessed-cake:
@ermalkaleci, how do you offer to use `subwasm`?
As far as I know, `subwasm metadata` generates metadata in a JSON format only, but `subxt` expect metadata in a binary format.

PR #85, opened 2022-12-13
support upgrade

ermalkaleci:
support to create custom proof, i.e: upgradeGoAhead signal to support upgrade


executor/src/lib.rs
```diff
@@ -2,17 +2,19 @@ extern crate console_error_panic_hook;
...
 mod task;
 use crate::runner_api::RpcApiClient;
-use smoldot::json_rpc::methods::HexString;
+use smoldot::json_rpc::methods::{HashHexString, HexString};
 
 #[wasm_bindgen]
 pub async fn get_metadata(code: &str) -> Result<JsValue, JsValue> {
     std::panic::set_hook(Box::new(console_error_panic_hook::hook));
-    _ = console_log::init_with_level(log::Level::Debug);
+    // _ = console_log::init_with_level(log::Level::Debug);
```
xlc, line 17: do we still need those or not?

ermalkaleci: we're not using it so I commented them in cause we use it later

xlc: can we just leave it on?

ermalkaleci: yes we can

PR #93, opened 2022-12-16
Wasm js bindings

ermalkaleci:


configs/genesis.yml
```diff
@@ -0,0 +1,13 @@
+genesis: https://raw.githubusercontent.com/AcalaNetwork/Acala/master/resources/mandala-dist.json
```
xlc, line 1: rename this file to `mandala-genesis.yml`

executor/Cargo.toml
```diff
@@ -1,6 +1,9 @@
 [package]
 name = "chopsticks-executor"
 version = "0.2.2"
+description = "Chopsticks executor"
+repository = "https://github.com/AcalaNetwork/chopsticks"
+license = ""
```
xlc, line 6:
```suggestion
license = "Apache-2.0"
```

executor/pkg/package.json
```diff
@@ -1,11 +1,17 @@
 {
   "name": "@acala-network/chopsticks-executor",
-  "version": "0.2.0",
+  "description": "Chopsticks executor",
+  "version": "0.2.2",
+  "license": "",
```
xlc, line 5:
```suggestion
  "license": "Apache-2.0",
```

src/bindings.ts
```diff
@@ -0,0 +1,29 @@
+import { HexString } from '@polkadot/util/types'
+
+import { Blockchain } from './blockchain'
+import { defaultLogger, truncate } from './logger'
+
+const logger = defaultLogger.child({ name: 'binding' })
+
+export const setupBindings = (chain: Blockchain) => {
+  global.getStorage = async function (blockHash: HexString, key: HexString) {
```
xlc, line 9: can we put those under a namespace to reduce the chance of conflicts?
something like `global._chopsticks_binding_ = { getStorage() {}, getPrefixKeys() {} }`

executor/src/lib.rs
```diff
@@ -1,67 +1,60 @@
...
+    {
+        let _ = console_log::init_with_level(log::Level::Trace);
+    }
 }
 
 #[wasm_bindgen]
-pub async fn get_runtime_version(code: &str) -> Result<JsValue, JsValue> {
-    std::panic::set_hook(Box::new(console_error_panic_hook::hook));
-    _ = console_log::init_with_level(log::Level::Debug);
+pub async fn get_runtime_version(code: JsValue) -> Result<JsValue, JsError> {
```
xlc, line 20: could we just pass a callback object for query storage instead of using global variables? otherwise it may be hard to support running multiple networks in a same nodejs process

xlc (changes requested):

executor/src/lib.rs
ermalkaleci: If we're going to run multiple nodes in the same process then one solution will be to have an id for the chain and pass that around 
> could we just pass a callback object for query storage instead of using global variables? otherwise it may be hard to support running multiple networks in a same nodejs process

src/bindings.ts
ermalkaleci: Yes good idea

executor/src/lib.rs
xlc: that will also work. it is I always want to avoid global variable if possible

ermalkaleci: I think callback is possible as well. Will see which is best

ermalkaleci: will address into different PR

PR #112, opened 2022-12-21
fix slot & timestamp

ermalkaleci:


src/setup.ts
```diff
@@ -45,12 +45,10 @@ export const setup = async (argv: Config) => {
 
   const header = await api.getHeader(blockHash)
 
-  const blockNumber = +header.number
-  const timestamp = argv.timestamp ?? Date.now()
-  const setTimestamp = new SetTimestamp((newBlockNumber) => {
-    return timestamp + (newBlockNumber - blockNumber) * 12000 // TODO: make this more flexible
-  })
+  // TODO: do we really need to set a custom timestamp?
+  // const timestamp = argv.timestamp
```
xlc, line 49: yes we do. some of Acala honzon logic depends on timestamp and we want to be able to mock it

ermalkaleci: in that case we need to have a helper method that modify slot and timestamp storage

ermalkaleci: SetTimestamp doesn't have to know anything. It will just produce valid timestamps

src/blockchain/inherent/index.ts
```diff
@@ -1,31 +1,27 @@
...
-  }
-
-  getTimestamp(blockNumber: number): number {
-    return this.#getTimestamp(blockNumber)
+  async createInherents(parent: Block): Promise<HexString[]> {
+    const meta = await parent.meta
+    const timestampRaw = (await parent.get(compactHex(meta.query.timestamp.now()))) || '0x'
+    const currentTimestamp = meta.registry.createType('u64', hexToU8a(timestampRaw)).toNumber()
+    const period = meta.consts.babe
+      ? (meta.consts.babe.expectedBlockTime.toJSON() as number)
```
xlc, line 21: why toJSON instead of toNumber?

src/setup.ts
xlc: I guess we could have a SetExactTimestamp or something and allow config to choose what inherents to include

src/blockchain/inherent/index.ts
ermalkaleci: because the type is Codec

src/setup.ts
ermalkaleci: it won't work will latest upgrade,  that's why I updated this

xlc (approved):

PR #153, opened 2023-01-20
View function poc - do not merge

shaunxw:


executor/src/task.rs
```diff
@@ -101,6 +102,8 @@ pub async fn run_task(task: TaskCall, js: crate::JsCallback) -> Result<TaskRespo
     .unwrap();
     let mut ret: Result<Vec<u8>, String> = Ok(Vec::new());
 
+	let mut keys: Vec<HexString> = Vec::new();
```
xlc, line 105: need to deupe this

src/nonce.ts
```diff
@@ -0,0 +1,48 @@
...
+      result: registry.createType('u64', hexToU8a(result.Call.result)).toNumber(),
+      accessedStorageKeys: result.Call.accessedStorageKeys,
+    }
+  }
+
+  const { accessedStorageKeys } = await getNonce()
+
+  const api = await ApiPromise.create({ provider: context.api.provider })
+  const unsub = await api.rpc.state.subscribeStorage(accessedStorageKeys, async (_changes) => {
+    const nonce = await getNonce();
```
xlc, line 43: need to resub the new storages just in case it is accessing something else

xlc (commented):
need to think about how to deal with next key and prefix keys

src/nonce.ts
```diff
@@ -0,0 +1,48 @@
...
+  const getNonce = async () => {
+    const result = await runTask(
+      {
+        wasm,
+        calls: [['AccountNonceApi_account_nonce', accountId]],
+        storage: [],
+        mockSignatureHost: false,
+        allowUnresolvedImports: false,
+      },
+      taskHandler(parent)
```
ermalkaleci, line 26: this is not correct. this need to be head

executor/src/task.rs
```diff
@@ -121,6 +124,9 @@ pub async fn run_task(task: TaskCall, js: crate::JsCallback) -> Result<TaskRespo
                 }
                 RuntimeHostVm::StorageGet(req) => {
                     let key = HexString(req.key().as_ref().to_vec());
+
+					keys.push(key.clone());
```
ermalkaleci, line 128: you don't need this. storage_diff has all the keys

src/nonce.ts
ermalkaleci: this can be `context.chain.head.call('AccountNonceApi_account_nonce', accountId)`

executor/src/task.rs
shaunxw: for getters storage_diff is always empty

PR #193, opened 2023-02-19
change to use the monorepo

ntduan:


e2e/src/import-storage/index.test.ts
```diff
@@ -2,7 +2,7 @@ import { describe, expect, it } from 'vitest'
 import path from 'path'
 
 import { api, chain, setupApi } from '../helper'
-import { importStorage, overrideWasm } from '../../src/utils/import-storage'
+import { importStorage, overrideWasm } from '@acala-network/chopsticks/lib/utils/import-storage'
```
ermalkaleci, line 5:
```suggestion
import { importStorage, overrideWasm } from '@acala-network/chopsticks/src/utils/import-storage'
```

It will be best if we run tests without having to build first

ermalkaleci:
is there any reason why we need tsconfig.json and tsconfig.build.json?

ermalkaleci:
please move template dir inside chopsticks

ntduan:
> is there any reason why we need tsconfig.json and tsconfig.build.json?

It is to avoid having the test files published together.

PR #206, opened 2023-02-27
improve port handling

xlc:
`setupWithServer` should not deal with env as it could be used as a library function

port could be zero which means let system to pick available port.


packages/chopsticks/src/cli.ts
```diff
@@ -26,8 +26,9 @@ const processConfig = async (path: string) => {
 
 const processArgv = async (argv: any) => {
   if (argv.config) {
-    return { ...(await processConfig(argv.config)), ...argv }
+    argv = { ...(await processConfig(argv.config)), ...argv }
   }
+  argv.port = argv.port ?? (process.env.PORT ? Number(process.env.PORT) : 8000)
```
xlc, line 31: cli overrides config file overrides env. maybe we want env to be top priority?

ermalkaleci (approved):

Issue #70, opened 2022-12-03
try-runtime integration


xlc:
New config to call `try_state` on every new block to assert integrality of the new block

PR #222, opened 2023-03-03
try runtime

dzmitry-lahoda:
may be this https://github.com/AcalaNetwork/chopsticks/issues/70 (so only None , no checks filter)

### DIFF
--- a/packages/chopsticks/src/cli.ts
+++ b/packages/chopsticks/src/cli.ts
@@ -82,18 +82,12 @@ yargs(hideBin(process.argv))
       await runBlock(await processArgv(argv))
     }
   )
-yargs(hideBin(process.argv))
-  .scriptName('chopsticks')
   .command(
     'try-runtime',
     'Runs runtime upgrade',
     (yargs) =>
       yargs.options({
         ...defaultOptions,
-        port: {
-          desc: 'Port to listen on',
-          number: true,
-        },
         'output-path': {
           desc: 'File path to print output',
           string: true,
--- a/packages/chopsticks/src/cli.ts
+++ b/packages/chopsticks/src/cli.ts
@@ -12,6 +12,7 @@ import { dryRun } from './dry-run'
 import { dryRunPreimage } from './dry-run-preimage'
 import { isUrl } from './utils'
 import { runBlock } from './run-block'
+import { tryRuntime } from './try-runtime'
 
 const processConfig = async (path: string) => {
   let file: string
@@ -79,6 +80,33 @@ yargs(hideBin(process.argv))
       }),
     async (argv) => {
       await runBlock(await processArgv(argv))
+    }
+  )
+yargs(hideBin(process.argv))
+  .scriptName('chopsticks')
+  .command(
+    'try-runtime',
+    'Runs runtime upgrade',
+    (yargs) =>
+      yargs.options({
+        ...defaultOptions,
+        port: {
+          desc: 'Port to listen on',
+          number: true,
+        },
+        'output-path': {
+          desc: 'File path to print output',
+          string: true,
+        },
+        html: {
+          desc: 'Generate html with storage diff',
+        },
+        open: {
+          desc: 'Open generated html',
+        },
+      }),
+    async (argv) => {
+      await tryRuntime(await processArgv(argv))
     }
   )
   .command(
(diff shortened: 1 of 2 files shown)


packages/chopsticks/src/try-runtime.ts
```diff
@@ -0,0 +1,55 @@
...
+import { generateHtmlDiffPreviewFile } from './utils/generate-html-diff'
+import { openHtml } from './utils/open-html'
+import { runTask, taskHandler } from './executor'
+import { setup } from './setup'
+
+export const tryRuntime = async (argv: Config) => {
+  const context = await setup(argv)
+  const wasm = await context.chain.head.wasm
+  const block = context.chain.head
+  const parent = await block.parentBlock
```
ermalkaleci, line 14: why do you need parent?

packages/chopsticks/src/cli.ts
```diff
@@ -81,6 +82,33 @@ yargs(hideBin(process.argv))
       await runBlock(await processArgv(argv))
     }
   )
+yargs(hideBin(process.argv))
```
ermalkaleci, line 85: you don't need those 2 lines
```suggestion
```

packages/chopsticks/src/cli.ts
```diff
@@ -81,6 +82,33 @@ yargs(hideBin(process.argv))
...
+  .scriptName('chopsticks')
+  .command(
+    'try-runtime',
+    'Runs runtime upgrade',
+    (yargs) =>
+      yargs.options({
+        ...defaultOptions,
```
ermalkaleci, line 92: we need to make wasm-override required for this command

packages/chopsticks/src/cli.ts
```diff
@@ -81,6 +82,33 @@ yargs(hideBin(process.argv))
...
+        port: {
```
ermalkaleci, line 93: no port is required

packages/chopsticks/src/try-runtime.ts
```diff
@@ -0,0 +1,55 @@
...
+    UpgradeCheckSelect: {
+      _enum: {
+        None: null,
+      },
+    },
+  })
+
+  const select_none = registry.createType('UpgradeCheckSelect', 'None')
+  const calls: [string, HexString[]][] = [['TryRuntime_on_runtime_upgrade', [select_none.toHex()]]]
+  const result = await runTask(
```
ermalkaleci, line 27:
```suggestion
  const result = await block.call('TryRuntime_on_runtime_upgrade', [select_none.toHex()])
```

packages/chopsticks/src/cli.ts
dzmitry-lahoda: i am not sure about it . in theory i can add data to chain and try runtime. without chaning wasm.

ermalkaleci: But you need to provide a wasm build with feature "try-runtime" right?

packages/chopsticks/src/cli.ts
```diff
@@ -81,6 +82,27 @@ yargs(hideBin(process.argv))
       await runBlock(await processArgv(argv))
     }
   )
+  .command(
+    'try-runtime',
+    'Runs runtime upgrade',
+    (yargs) =>
+      yargs.options({
+        ...defaultOptions,
+        'output-path': {
```
ermalkaleci, line 91: add it here please
```suggestion
'wasm-override': {
    desc: '...',
    string: true,
    required: true
  },
        'output-path': {
```

dzmitry-lahoda: so it will not check override in config? so we have 2 places of checks. is 2 checks okey?

ermalkaleci: I mean remove check from tryRuntime function and add it here. `wasm-override is` defined on `...defaultOptions` but it's optional. we are making it required for try-runtime command

ermalkaleci: Having it on cli command will give you the benefit of help `yarn start try-runtime --help`

ermalkaleci: @dzmitry-lahoda can you do this last change so we can merge?

PR #223, opened 2023-03-06
won't support sending tx with out of order nonce

xlc:
this made debugging e2e test very hard as bad nonce means it hangs forever


packages/chopsticks/src/blockchain/block-builder.ts
```diff
@@ -153,17 +154,45 @@ export const buildBlock = async (
...
+  }
+
   // apply extrinsics
   for (const extrinsic of extrinsics) {
     try {
       const { result, storageDiff } = await newBlock.call('BlockBuilder_apply_extrinsic', [extrinsic])
       const outcome = registry.createType<ApplyExtrinsicResult>('ApplyExtrinsicResult', result)
       if (outcome.isErr) {
-        if (outcome.asErr.isInvalid && outcome.asErr.asInvalid.isFuture) {
-          pendingExtrinsics.push(extrinsic)
```
ermalkaleci, line 163: why removing this?

xlc: it only makes tests harder to write. I don't see why we need this feature

PR #245, opened 2023-03-13
try more ports before giving up

sander2:
If connecting to > 5 chains, the current code fails to find ports for all of the chains. There is probably a better fix possible but as a quick fix this is fine


ermalkaleci:
You should specify a port. That 5 attempts is in case it is taken. If you want to start at a random port just use 0

PR #264, opened 2023-04-19
load remote configs by name

ermalkaleci:


packages/chopsticks/src/cli.ts
```diff
@@ -13,17 +14,34 @@ import { decodeKey } from './utils/decoder'
...
+const CONFIGS_BASE_URL = 'https://raw.githubusercontent.com/AcalaNetwork/chopsticks/master/configs/'
+
 const processConfig = async (path: string) => {
   let file: string
   if (isUrl(path)) {
     file = await axios.get(path).then((x) => x.data)
   } else {
-    file = readFileSync(path, 'utf8')
+    try {
+      accessSync(path, constants.R_OK)
```
xlc, line 31: why is this needed?

ermalkaleci: Make it easy to load some default configs. We can make configs to be more generic

ermalkaleci: Do you mean accessSync or this PR?

ermalkaleci: It's just checking if can access that path, not super important

ermalkaleci: Yeah I was trying something else and left this there

xlc (approved):

PR #317, opened 2023-06-08
a few config changes

goncer:
* Add more tokens in mangata config.
* Update endpoint
* Log level to get node logs
* Create bifrost.yml
* Create imbu.yml

### DIFF
--- a/configs/bifrost.yml
+++ b/configs/bifrost.yml
@@ -9,5 +9,6 @@ import-storage:
       -
         -
           - 5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY
-        - data:
+        - providers: 1
+          data:
             free: 1000000000000000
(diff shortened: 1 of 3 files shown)


xlc (approved):

configs/bifrost.yml
```diff
@@ -0,0 +1,13 @@
...
+block: ${env.BIFROST_BLOCK_NUMBER}
+db: ./db.sqlite
+
+import-storage:
+  System:
+    Account:
+      -
+        -
+          - 5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY
+        - data:
```
ermalkaleci, line 12: @goncer please set providers to 1 so you don't have to updated this later

goncer: what do you mean by providers? not sure I understand you :upside_down_face:

goncer: updated :crossed_fingers:

ermalkaleci: yes, latest substrate version requires providers so when you upgrade it will still work

PR #331, opened 2023-07-09
feat: respect system proxy

Misaka-0x447f:
Some of api endpoints may not available in China mainland and this feature can help.


xlc (commented):
global-tunnel-ng lookes like old, how about https://www.npmjs.com/package/global-agent or something newer?

README.md
```diff
@@ -52,6 +52,8 @@ Make sure you have setup Rust environment (>= 1.64).
 - Build wasm
   - `yarn build-wasm`
 
+For webstorm users, you may need to run `yarn build-wasm` in terminal rather than webstorm's build-in run tool.
```
ermalkaleci, line 55: this is not the case. we doing local setup you need to build wasm always, it's mention above

packages/chopsticks/src/setup.ts
```diff
@@ -1,13 +1,11 @@
...
 import '@polkadot/types-codec'
-import { DataSource } from 'typeorm'
-import { HexString } from '@polkadot/util/types'
-import { ProviderInterface } from '@polkadot/rpc-provider/types'
-import { WsProvider } from '@polkadot/api'
-
 import { Api } from './api'
 import { Blockchain } from './blockchain'
 import { Config } from './schema'
+import { DataSource } from 'typeorm'
```
xlc, line 6: why move those?

xlc (approved):

PR #341, opened 2023-07-17
Adding support for Interlay, Tinkernet, and update to the HydraDX configuration

x676f64:
I've added support for Interlay, Tinkernet, and included balances for all defined assets. I've also extended HydraDX's config to include all assets there.

### DIFF
--- a/configs/hydradx.yml
+++ b/configs/hydradx.yml
@@ -5,6 +5,14 @@ db: ./hydradx.db.sqlite
 # wasm-override: hydradx_runtime.compact.compressed.wasm
 
 import-storage:
+  System:
+    Account:
+      -
+        -
+          - 5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY
+        - providers: 1
+          data:
+            free: 1000000000000000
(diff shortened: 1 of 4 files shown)


configs/hydradx.yml
```diff
@@ -1,22 +1,99 @@
 endpoint: wss://hydradx-rpc.dwellir.com
 mock-signature-host: true
 block: ${env.HYDRADX_BLOCK_NUMBER}
-db: ./db.sqlite
+db: ./hydradx.db.sqlite
 # wasm-override: hydradx_runtime.compact.compressed.wasm
 
 import-storage:
-  System:
```
ermalkaleci, line 8: no need for system::account?

arrudagates: Interlay uses Tokens for both native and external assets.

x676f64: Thanks, this is needed. Added it back.

ermalkaleci:
@x676f64 let me know when it's ready to merge

x676f64:
> 



> @x676f64 let me know when it's ready to merge

Should be good to merge now

PR #346, opened 2023-07-24
no need to use the At rpc

xlc:
they are alias so no need to distinguish them


packages/chopsticks/src/api.ts
```diff
@@ -103,16 +103,10 @@ export class Api {
   }
 
   async getStorage(key: string, hash?: string) {
-    if (hash) {
-      return this.#provider.send<string>('state_getStorageAt', [key, hash])
-    }
-    return this.#provider.send<string>('state_getStorage', [key])
+    return this.#provider.send<string>('state_getStorage', [key, hash])
```
ermalkaleci, line 106:
```suggestion
    const params = [key]
    if (hash) params.push(hash)
    return this.#provider.send<string>('state_getStorage', params)
```

packages/chopsticks/src/api.ts
```diff
@@ -103,16 +103,10 @@ export class Api {
...
   }
 
   async getKeysPaged(prefix: string, pageSize: number, startKey: string, hash?: string) {
-    if (hash) {
-      return this.#provider.send<string[]>('state_getKeysPagedAt', [prefix, pageSize, startKey, hash])
-    }
-    return this.#provider.send<string[]>('state_getKeysPaged', [prefix, pageSize, startKey])
+    return this.#provider.send<string[]>('state_getKeysPaged', [prefix, pageSize, startKey, hash])
```
ermalkaleci, line 110:
```suggestion
    const params = [prefix, pageSize, startKey]
    if (hash) params.push(hash)
    return this.#provider.send<string[]>('state_getKeysPaged', params)
```

Issue #347, opened 2023-07-27
Persistent mode


xlc:
- [ ] Save chain/blocks to sqlite db
- [ ] Load chain from db
- [ ] Config to specify max number of blocks to be stored in memory. If db is enabled, save them to db before purge.

qiweiii:
- [ ] Config to specify max number of blocks to be stored in memory.
  - background: problem is now all blocks are in memory, too many blocks will break the program.
  - [ ] stage 1: when exceed the max number of block, delete old blocks, so it won't break
  - [ ] stage 1: If db is enabled, save old blocks to db before purge
- [ ] Save chain/blocks to sqlite db before purge
- [ ] Load chain/blocks from db, and resume from where it stopped

xlc:
No need to change build block mode. We still want this with all the existing modes.

qiweiii:
updated

PR #353, opened 2023-08-01
Max memory block count

qiweiii:
solves part of #347

### DIFF
--- a/packages/chopsticks/src/blockchain/index.ts
+++ b/packages/chopsticks/src/blockchain/index.ts
@@ -30,6 +30,7 @@ export interface Options {
   runtimeLogLevel?: number
   registeredTypes: RegisteredTypes
   offchainWorker?: boolean
+  maxMemoryBlockCount?: number
 }
 
 export class Blockchain {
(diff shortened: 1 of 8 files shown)


packages/chopsticks/src/blockchain/index.ts
```diff
@@ -83,10 +86,21 @@ export class Blockchain {
...
       this.offchainWorker = new OffchainWorker()
     }
+
+    this.#maxMemoryBlockCount = maxMemoryBlockCount
   }
 
   #registerBlock(block: Block) {
-    this.#blocksByNumber[block.number] = block
+    // refresh key
+    if (this.#blocksByNumber.has(block.number)) {
```
xlc, line 95: why need to delete this first?

qiweiii: I was trying to make it like an lru cache. `getBlockAt` and `getBlock` also calls `#registerBlock`, if right now block 10 to 20 are in memory, and user getBlock(10), that means they need to access it, so I re-insert it into the `Map`, so the first one to be deleted will be block 11, and block 10 is still in memory for faster access

If we always want to delete the smallest block number, then we don't need this.

qiweiii: wait, I was wrong, `#registerBlock` is only called when block is not found, so we actually don't need this😅

xlc (approved):

PR #359, opened 2023-08-02
Save blocks to db

qiweiii:
related to #347 
- [x] need to merge #353 first
- [x] add tests

image


packages/chopsticks/src/blockchain/index.ts
```diff
@@ -125,9 +140,18 @@ export class Blockchain {
...
         const loadingBlock = (async () => {
           try {
-            const header = await this.api.getHeader(hash)
-            const block = new Block(this, Number(header.number), hash)
-            this.#registerBlock(block)
+            if (this.db) {
+              const blockData = await this.db.getRepository('Block').findOne({ where: { hash } })
+              if (blockData) {
+                const { number, hash, header, extrinsics } = blockData
+                const block = new Block(this, number, hash, undefined, { header, extrinsics })
```
qiweiii, line 147: I guess we don't need to apply extrinsics and get storageDiff here

qiweiii: is it necessary to add parent block here?

packages/chopsticks/src/blockchain/index.ts
```diff
@@ -83,11 +86,23 @@ export class Blockchain {
...
+  async #registerBlock(block: Block) {
+    // if exceed max memory block count, delete the oldest block
+    if (this.#blocksByNumber.size === this.#maxMemoryBlockCount) {
+      const firstKey = this.#blocksByNumber.keys().next().value
+      this.#blocksByNumber.delete(firstKey)
+    }
+    this.#blocksByNumber.set(block.number, block)
     this.#blocksByHash[block.hash] = block
+    // save to db
+    if (this.db) {
```
qiweiii, line 102: is it more reasonable to save on close server or save on register block? i think save here is easier and also may have better user experience

packages/chopsticks/src/blockchain/index.ts
```diff
@@ -105,12 +120,12 @@ export class Blockchain {
     if (number > this.#head.number) {
       return undefined
     }
-    if (!this.#blocksByNumber[number]) {
+    if (!this.#blocksByNumber.has(number)) {
       const hash = await this.api.getBlockHash(number)
```
qiweiii, line 124: I didn't get from db in getBlockAt as the hash is not provided, if get from db by number then it could be reliable if there are duplicated blocks

packages/chopsticks/src/blockchain/index.ts
xlc: Save on register is fine. Otherwise we may have long shutdown time or lost data if server is forced killed.

packages/chopsticks/src/blockchain/index.ts
xlc: in that case we need a way to mark a block as canonical. one easy way is delete the old one. alternatively mark the new one as canonical and unmark the old canonical block

packages/chopsticks/src/blockchain/index.ts
xlc: we need parent block.
I think we need to save storageDiff as for example it is possible to use `dev_setStorage` to override them and without saving those, the chain will be different after restart, which is unexpected.

packages/chopsticks/src/blockchain/index.ts
```diff
@@ -90,14 +91,18 @@ export class Blockchain {
...
+  async #registerBlock(block: Block, saveToDB = true) {
     // if exceed max memory block count, delete the oldest block
     if (this.#blocksByNumber.size === this.#maxMemoryBlockCount) {
       const firstKey = this.#blocksByNumber.keys().next().value
       this.#blocksByNumber.delete(firstKey)
     }
     this.#blocksByNumber.set(block.number, block)
     this.#blocksByHash[block.hash] = block
+    // save to db
+    if (saveToDB) {
```
ermalkaleci, line 103: can you remove this from here and call it only when needed?

packages/chopsticks/src/blockchain/index.ts
```diff
@@ -108,6 +113,44 @@ export class Blockchain {
...
+        },
+        ['hash'],
+      )
+    }
+  }
+
+  /**
+   * Try to get block from db, if pass in number, get block by number, else get block by hash
+   */
+  async #getBlockFromDB(key: number | HexString): Promise<Block | undefined> {
```
ermalkaleci, line 138: either make this `loadBlockFromDB` or remove `registerBlock` so a get function doesn't have side effects

packages/chopsticks/src/blockchain/index.ts
qiweiii: any specific reason? I actually want to save block when register, otherwise this logic would be sacttered around

packages/chopsticks/src/blockchain/index.ts
qiweiii: sure, let's rename it `loadBlockFromDB`

packages/chopsticks/src/blockchain/index.ts
ermalkaleci: not every block should go to DB, only those we create

ermalkaleci: setHead for example should save block

ermalkaleci:
@xlc @qiweiii this isn't making much sense to me. We need to save block on db for caching, not every block fetched/created. Basically when block is removed from memory it is saved into db for later use

qiweiii:
> @xlc @qiweiii this isn't making much sense to me. We need to save block on db for caching, not every block fetched/created. Basically when block is removed from memory it is saved into db for later use

right, maybe my understanding is wrong, if we want to resume from where we last stopped, is it necessary to save every block? If not, I guess we don't need to save them

xlc:
We want to save all the new blocks. We don't need to save blocks fetched from remote API. Right now it is saving all the blocks including the one fetched from remote API.

ermalkaleci:
should work like this

chain: 1 -> 2 -> 3  -> 4 -> 5 -> head 6
                 fork    |
     chopsticks.  3(memory only) -> 4(memory+db) -> head 5(memory+db)



chopsticks start with --resume flag then load head from db
   chopsticks. 2 (db ?? api) -> 3 (db ?? api) -> 4(db ?? api) -> head 5(memory+db)

packages/chopsticks/src/blockchain/block-builder.ts
```diff
@@ -135,6 +135,8 @@ const initNewBlock = async (head: Block, header: Header, inherents: HexString[],
     }
   }
 
+  await head.chain.saveBlockToDB(newBlock)
```
ermalkaleci, line 138: this is wrong here

packages/chopsticks/src/blockchain/block-builder.ts
```diff
@@ -287,6 +289,8 @@ export const buildBlock = async (
     storageDiff,
   })
 
+  await head.chain.saveBlockToDB(finalBlock)
```
ermalkaleci, line 292: this shouldn't be here. create a method on Blockchain.class `onNewBlock` which deals with new blocks and replace txpool `await this.#chain.setHead(newBlock)` with `await this.#chain.onNewBlock(newBlock)`

qiweiii:
```diff
+ async onNewBlock(block: Block): Promise<void> {
+   await this.setHead(block);
+   await this.saveBlockToDB(block);
+ }
```

do you mean adding a `onNewBlock` method like this?

ermalkaleci (approved):

packages/chopsticks/src/blockchain/index.ts
```diff
@@ -108,6 +109,45 @@ export class Blockchain {
     return this.#txpool
   }
 
+  async saveBlockToDB(block: Block) {
+    if (this.db) {
+      const { hash, number, header, extrinsics } = block
+      // delete old ones with the same block number if any, keep the latest one
+      await this.db.getRepository(BlockEntity).delete({ number })
+      await this.db.getRepository(BlockEntity).upsert(
```
xlc, line 117: should be transactional to avoid partial state update

packages/chopsticks/src/blockchain/index.ts
```diff
@@ -162,6 +209,15 @@ export class Blockchain {
       this.#blocksByNumber.delete(block.number)
     }
     delete this.#blocksByHash[block.hash]
+    // delete from db
+    if (this.db) {
+      this.db.getRepository(BlockEntity).delete({ hash: block.hash })
```
xlc, line 214: should await this

packages/e2e/src/blocks-save.test.ts
```diff
@@ -0,0 +1,27 @@
...
+    const numberOfBlocks = await chain.db?.getRepository('Block').count()
+    expect(numberOfBlocks).toEqual(2)
+
+    const block = await chain.getBlockAt(chain.head.number)
+    const blockData = await chain.db?.getRepository('Block').findOne({ where: { number: chain.head.number } })
+    expect(blockData?.hash).toEqual(block?.hash)
+    expect(JSON.stringify(blockData?.header)).toEqual(JSON.stringify(block?.header))
+    expect(blockData?.parentHash).toEqual((await block?.parentBlock)?.hash)
+    expect(JSON.stringify(blockData?.extrinsics)).toEqual(JSON.stringify(await block?.extrinsics))
+    expect(JSON.stringify(blockData?.storageDiff)).toEqual(JSON.stringify(await block?.storageDiff()))
```
xlc, line 25: should avoid `?` in tests. if `blockData` and `blocks` are undefined, the test will pass.
use `!` to ensure it will error if the value is undefined

xlc:
I think with this, we will be able to load chain from db by specifying right block hash. Need a manual test and e2e test to confirm that.

packages/e2e/src/blocks-save.test.ts
```diff
@@ -13,15 +13,15 @@ describe('block-save', async () => {
...
     expect(numberOfBlocks).toEqual(2)
 
     const block = await chain.getBlockAt(chain.head.number)
-    const blockData = await chain.db?.getRepository('Block').findOne({ where: { number: chain.head.number } })
-    expect(blockData?.hash).toEqual(block?.hash)
-    expect(JSON.stringify(blockData?.header)).toEqual(JSON.stringify(block?.header))
-    expect(blockData?.parentHash).toEqual((await block?.parentBlock)?.hash)
-    expect(JSON.stringify(blockData?.extrinsics)).toEqual(JSON.stringify(await block?.extrinsics))
-    expect(JSON.stringify(blockData?.storageDiff)).toEqual(JSON.stringify(await block?.storageDiff()))
+    const blockData = await chain.db!.getRepository('Block').findOne({ where: { number: chain.head.number } })
```
xlc, line 20: should check `block` and `blockData` is not undefined and no need all the `!`

qiweiii:
> I think with this, we will be able to load chain from db by specifying right block hash. Need a manual test and e2e test to confirm that.

Maybe do this in a separate PR coz may need to change `setup` for load to work properly?

xlc (approved):

PR #373, opened 2023-08-18
Fix api

ermalkaleci:


packages/core/src/blockchain/storage-layer.ts
```diff
@@ -42,13 +42,13 @@ export class RemoteStorageLayer implements StorageLayerProvider {
     if (this.#db) {
       const res = await keyValuePair?.findOne({ where: { key, blockHash: this.#at } })
       if (res) {
-        return res.value
+        return res.value ?? undefined
```
xlc, line 45: why `?? undefined`? that's a nop

ermalkaleci: Db returns null

PR #379, opened 2023-08-24
export `signFake` helper

### LINKED ISSUE #378: export `fakeSign`
currently when we want to fake sign a chopsticks tx, we need to manually add this helper
```ts
  const fakeSign = (tx, addr, nonce) => {
    const mockSignature = new Uint8Array(64)
    mockSignature.fill(0xcd)
    mockSignature.set([0xde, 0xad, 0xbe, 0xef])
    tx.signFake(addr, {
      nonce,
      genesisHash: api.genesisHash,
      runtimeVersion: api.runtimeVersion,
      blockHash: api.genesisHash,
    })
    // update fake signature
    tx.signature.set(mockSignature)

    return tx
  }

    const tx = api.tx.xxx()
    fakeSign(tx, testaddr, 0)
 ```

As a developer, I don't care about dead beaf or deaf wolf, I just want a fake sign. So it would be nice if we can just do
```ts
import { fakeSign } from 'chopstciks/utils'
const tx = api.tx.xxx()
fakeSign(tx, testaddr, 0)
```

it's even more conveninet if we don't need to manually compute the nonce
```ts
fakeSign(api, tx, testaddr, { ...overrides })   // computes correct nonce internally if no nonce override 
```

shunjizhan:
## Change
fix #378 

- exported 2 helpers:
  - `signFakeWithApi()`: use api to auto fetch info
  - `signFake()`: manually input sign options
- imported polkadot type augment, so compiler can infer correct type for `api.query.system.account`
- exported utils from index so devs can import from `@acala-network/chopsticks/utils`

## Test
modified previous mock signature tests to test these 2 helpers, still good


packages/chopsticks/src/utils/signFake.ts
```diff
@@ -0,0 +1,37 @@
+import { ApiPromise } from "@polkadot/api";
+import { GenericExtrinsic } from "@polkadot/types"
+import { SignatureOptions } from "@polkadot/types/types"
+
+export type signFakeOptions = Partial<SignatureOptions>;
```
xlc, line 5:
```suggestion
export type SignFakeOptions = Partial<SignatureOptions>;
```

xlc (approved):

Issue #363, opened 2023-08-10
dev rpc to set runtime log level


xlc:
useful for debugging

PR #390, opened 2023-09-05
allow dev rpc set runtime log level

qiweiii:
closes: https://github.com/AcalaNetwork/chopsticks/issues/363

allow user to set runtime log level, useful for debugging


packages/core/src/setup.ts
```diff
@@ -36,6 +37,9 @@ type Options = {
 }
 
 export const setup = async (options: Options) => {
+  defaultLogger.level = getLogLevel(options.runtimeLogLevel)
```
ermalkaleci, line 40: this has nothing to do with chopsticks logger. this is runtime logger

ermalkaleci: it is set here https://github.com/AcalaNetwork/chopsticks/pull/390/files#diff-42b25a5fdf620c3a1f0a17037770d594bb96251313e608d51a60eac9f2a6637eR109
This features should add a RPC call to change this value

qiweiii: got it, @xlc has messaged me as well

packages/chopsticks/src/plugins/set-runtime-log-level/index.ts
```diff
@@ -0,0 +1,14 @@
+import { Handler, ResponseError } from '../../rpc/shared'
+import { defaultLogger } from '../../logger'
+
+export const rpc: Handler = async (context, [runtimeLogLevel]) => {
+  defaultLogger.debug({ runtimeLogLevel }, 'dev_setRuntimeLogLevel')
+
+  if (typeof runtimeLogLevel !== 'number' || runtimeLogLevel < 0 || runtimeLogLevel > 5) {
```
ermalkaleci, line 7: no need to check the value, leave it to user to set correct value because it depends on runtime which value is correct

packages/chopsticks/src/plugins/set-runtime-log-level/index.ts
```diff
@@ -0,0 +1,14 @@
...
+    throw new ResponseError(1, `Invalid runtimeLogLevel ${runtimeLogLevel}`)
+  }
+
+  context.chain.runtimeLogLevel = runtimeLogLevel
+
+  defaultLogger.debug(`Runtime log level set to ${runtimeLogLevel}`)
```
xlc, line 13: the better place to log is in the setter

xlc (approved):

PR #392, opened 2023-09-08
feat: load blocks from db

qiweiii:
closes: #347

### DIFF
--- a/packages/chopsticks/src/cli.ts
+++ b/packages/chopsticks/src/cli.ts
@@ -46,6 +46,7 @@ const commands = yargs(hideBin(process.argv))
         },
         resume: {
           desc: 'Resume from the lastest block saved in the db, note this will override the block option',
+          boolean: true,
         },
       }),
     async (argv) => {
--- a/packages/chopsticks/src/cli.ts
+++ b/packages/chopsticks/src/cli.ts
@@ -45,8 +45,10 @@ const commands = yargs(hideBin(process.argv))
           number: true,
         },
         resume: {
-          desc: 'Resume from the lastest block saved in the db, note this will override the block option',
-          boolean: true,
+          desc: `Resume from the specified block hash or block number in db.
+                 If true, it will resume from the latest block in db.
+                 Note this will override the block option`,
+          string: true,
         },
       }),
     async (argv) => {
--- a/packages/chopsticks/src/cli.ts
+++ b/packages/chopsticks/src/cli.ts
@@ -44,6 +44,9 @@ const commands = yargs(hideBin(process.argv))
           desc: 'Max memory block count',
           number: true,
         },
+        resume: {
+          desc: 'Resume from the lastest block saved in the db, note this will override the block option',
+        },
       }),
     async (argv) => {
       await setupWithServer(argv as Config)
--- a/packages/chopsticks/src/context.ts
+++ b/packages/chopsticks/src/context.ts
@@ -23,13 +23,22 @@ export const setupContext = async (argv: Config, overrideParent = false) => {
   // load block from db
   if (chain.db) {
     if (argv.resume) {
-      const blockData = await chain.db.getRepository(BlockEntity).findOne({ where: {}, order: { number: 'desc' } })
+      const where: Record<string, string | number> = {}
+      switch (typeof argv.resume) {
+        case 'string':
+          where.hash = argv.resume
+          break
+        case 'number':
+          where.number = argv.resume
+          break
+        default:
+          break
+      }
+      const blockData = await chain.db.getRepository(BlockEntity).findOne({ where, order: { number: 'desc' } })
       if (blockData) {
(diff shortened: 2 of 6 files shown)


packages/chopsticks/src/context.ts
```diff
@@ -34,5 +35,15 @@ export const setupContext = async (argv: Config, overrideParent = false) => {
   await overrideStorage(chain, argv['import-storage'], at)
   await overrideWasm(chain, argv['wasm-override'], at)
 
+  // load blocks from db
```
qiweiii, line 38: is this the best place to load blocks? may need to consider do this before or after overrideStorage

packages/chopsticks/src/context.ts
```diff
@@ -34,5 +35,15 @@ export const setupContext = async (argv: Config, overrideParent = false) => {
...
+  if (chain.db && argv.resume) {
+    const blocks = await chain.db.getRepository(BlockEntity).find({ where: {}, order: { number: 'asc' } })
```
ermalkaleci, line 40: this is not enough. what happens if there're old blocks not related to our fork.
The easy way I can think is to resume if first db block is head+1 otherwise they must be considered invalid, also starting without resume should clear blocks from db and have a fresh start

packages/chopsticks/src/cli.ts
```diff
@@ -44,6 +44,9 @@ const commands = yargs(hideBin(process.argv))
           desc: 'Max memory block count',
           number: true,
         },
+        resume: {
```
ermalkaleci, line 47:
```suggestion
        resume: {
           boolean: true,
```

packages/chopsticks/src/context.ts
qiweiii: > what happens if there're old blocks not related to our fork

we can let user to make sure the db file they are using is the correct one, do you think this is possible?

> resume if first db block is head+1

for `head`, do you mean the `block` in cli option? if it defaults to the latest block, then we won't load blocks in db, is this what we want?

> starting without resume should clear blocks from db and have a fresh start

Yes, this make sense

ermalkaleci: > we can let user to make sure the db file they are using is the correct one, do you think this is possible?
> 
> >

No we can't because if I start once at block 100 and build 2 blocks then second time start at block 105 build 2 more blocks then I will have 101, 102, 106, 107 in database which are from same chain but not related

ermalkaleci: > for `head`, do you mean the `block` in cli option? if it defaults to the latest block, then we won't load blocks in db, is this what we want?

after chain is setup, `chain.head`

packages/e2e/src/blocks-save.test.ts
```diff
@@ -25,5 +24,19 @@ describe('block-save', async () => {
...
     expect(JSON.stringify(blockData.extrinsics)).toEqual(JSON.stringify(await block.extrinsics))
     expect(JSON.stringify(blockData.storageDiff)).toEqual(JSON.stringify(await block.storageDiff()))
+
+    savedBlockHash = block.hash
+
+    await acala.teardown()
+  })
+
+  it('load chain from the saved blocks', async () => {
+    const acala = await networks.acala({ db: resolve(tmpdir(), 'testdb.sqlite'), resume: true })
```
xlc, line 34: A test shouldn't depend on another test

packages/chopsticks/src/context.ts
xlc: Yeah I think load db before override so the overrides actually works.

packages/chopsticks/src/context.ts
```diff
@@ -19,6 +20,19 @@ export const setupContext = async (argv: Config, overrideParent = false) => {
...
+  // load block from db
+  if (chain.db) {
+    if (argv.resume) {
+      const blockData = await chain.db.getRepository(BlockEntity).findOne({ where: {}, order: { number: 'desc' } })
+      if (blockData) {
+        const block = await chain.loadBlockFromDB(blockData?.number)
+        block && (await chain.setHead(block))
+      }
+    } else {
+      await chain.db.getRepository(BlockEntity).clear()
```
xlc, line 32: why clear?

packages/chopsticks/src/cli.ts
```diff
@@ -44,6 +44,10 @@ const commands = yargs(hideBin(process.argv))
...
```
xlc, line 47: best to allow user to specify a block number or hash as well

packages/chopsticks/src/context.ts
qiweiii: to give them a fresh start? and also free some space

xlc: it is always bad to delete thing without user explicitly ask for it. we should have a purge command for that

xlc (approved):

Issue #376, opened 2023-08-22
TypeError: header.toJSON is not a function


xlc:
```
 Error handling request: TypeError: header.toJSON is not a function 'TypeError: header.toJSON is not a function
    at processHeader (/home/runner/acalajs-examples/node_modules/@acala-network/chopsticks/lib/rpc/substrate/chain.js:5:24)
    at chain_getHeader (/home/runner/acalajs-examples/node_modules/@acala-network/chopsticks/lib/rpc/substrate/chain.js:22:16)'
```

I think it has something to do with loading blocks from db

PR #393, opened 2023-09-08
fix header.toJSON is not a function

qiweiii:
closes: https://github.com/AcalaNetwork/chopsticks/issues/376

### DIFF
--- a/packages/chopsticks/src/rpc/substrate/chain.ts
+++ b/packages/chopsticks/src/rpc/substrate/chain.ts
@@ -1,12 +1,8 @@
 import { Handlers, ResponseError } from '../shared'
+import { Header } from '@polkadot/types/interfaces'
 
-const processHeader = (header: any) => {
-  let res: any
-  if (header?.toJSON) {
-    res = header.toJSON()
-  } else {
-    res = header
-  }
+const processHeader = (header: Header) => {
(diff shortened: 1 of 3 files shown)


packages/chopsticks/src/rpc/substrate/chain.ts
```diff
@@ -1,7 +1,12 @@
 import { Handlers, ResponseError } from '../shared'
 
 const processHeader = (header: any) => {
-  const res = header.toJSON() as any
```
ermalkaleci, line 4: this is not a proper fix. Header should have `toJSON` and be typed Header. I think the problem is where block is created, header is not encoded. One place is loadBlocksFromDb. It needs to encode header `registry.createType('Header', header)`

ermalkaleci:
```
export const BlockEntity = new EntitySchema<{
  hash: HexString
  number: number
  header: Header /// this is wrong, this is object
  parentHash: HexString | null
  extrinsics: HexString[]
  storageDiff: Record<HexString, HexString | null> | null
}>({
```

ermalkaleci: header type is wrong, needs to be object

ermalkaleci (approved):

PR #402, opened 2023-09-14
fix web test build

qiweiii:
For web-test, the current production bundle failed to perform db operations as typeorm package in browser has some dynamic `require` which are not transpiled, as discussed in the issue https://github.com/typeorm/typeorm/issues/8738

Using the suggested patch in that issue, I created this patch which works.

I didn't use `parcel` bundler as its production bundle also has problems that also need some research.


ermalkaleci (approved):
maybe in the future we use indexeddb directly on browser instead of sqljs

PR #409, opened 2023-09-21
add rpc types and docs

qiweiii:
add rpc param types and docs


packages/chopsticks/src/rpc/substrate/author.ts
```diff
@@ -1,11 +1,45 @@
...
+    subscriptionManager: SubscriptionManager,
+  ) => Promise<void>
+  author_pendingExtrinsics: (context: Context) => Promise<HexString[]>
+}
+
+/**
+ * Substrate `author` RPC methods, see {@link AuthorHandlers} for methods details.
+ */
+const handlers: AuthorHandlers = {
   author_submitExtrinsic: async (context, [extrinsic]) => {
```
ermalkaleci, line 43: @qiweiii  can't we add necessary types here?

ermalkaleci (approved):

PR #410, opened 2023-09-25
add core package docs

qiweiii:


ermalkaleci:
I don't think we need to have a comment for each method or property. Most of the cases they explain themselves

qiweiii:
Removed some obvious ones

xlc (approved):

PR #415, opened 2023-09-27
improve web test

### LINKED ISSUE #398: improve web test
- disable dry run call button until it is loaded
- allow user to input url and block number
- fill the call text field with a default call `0x0a000088dc3417d5058ec4b4503e0c12ea1a0a89be200fe98922423d4334014fa6b0ee0f0090c04bb6db2b`

qiweiii:
- closes #398 

- refactored into React code with mui components for styles

https://github.com/AcalaNetwork/chopsticks/assets/32790369/b8aafbb3-e4ac-4b7d-bb96-9db296565d07


ermalkaleci:
I prefer not complicating this as its purpose is to make sure it runs in browsers and not demo things. I think we should have another repo doing more complex features

qiweiii:
> I prefer not complicating this as its purpose is to make sure it runs in browsers and not demo things. I think we should have another repo doing more complex features

Got it, but I think this is still a very simple test case, and if we mention it in other [issues](https://github.com/polkadot-js/apps/issues/9903) it looks better. and it's good to test it inside modern framework since ppl are unlikely to use vanilla JS when using it in browser

ermalkaleci:
I don't want anyone to use this. This is only for testing

qiweiii:
By it, I mean chopsticks-core

xlc:
we do want to make this a demo. the demo should live with the same repo so that when we develop a new feature, we can dogfooding easily.

xlc (approved):

PR #430, opened 2023-10-10
fix browser undefined

ermalkaleci:


packages/core/src/blockchain/txpool.ts
```diff
@@ -238,20 +238,21 @@ export class TxPool {
...
-                case 'initialize':
-                  logger.trace(truncate(resp.storageDiff), 'Initialize block')
-                  break
-                case 'finalize':
-                  logger.trace(truncate(resp.storageDiff), 'Finalize block')
-                  break
-                default:
-                  logger.trace(truncate(resp.storageDiff), `Apply extrinsic ${phase}`)
+        onPhaseApplied:
+          logger.level.toLowerCase() === 'trace'
```
xlc, line 242: that's fine for trace, but not for other logging levels. still need to figure out why `isLevelEnabled` is not supported on browser

PR #432, opened 2023-10-11
Chopsticks provider

qiweiii:
related: https://github.com/polkadot-js/apps/issues/9903

A provider that can be used in ApiPromise, to be integrated here: https://github.com/polkadot-js/apps/compare/master...qiweiii:polkadot-js-apps:integrate-chopsticks

### Todo
- [x] fix and test tx and subscribe
- [ ] fix playwright tests (seems cannot work, posted a issue in playwright)
- [x] fix build
- [x] test build with polkadotjs app

https://github.com/AcalaNetwork/chopsticks/assets/32790369/e9ee6d6a-45e9-48d3-aaad-441b0a28aa8a


packages/core/src/chopsticks-provider.ts
```diff
@@ -0,0 +1,209 @@
...
+      throw new Error('ChopsticksProvider requires the upstream endpoint')
+    }
+    this.#endpoint = endpoint
+    // FIXME: WARNING in /node_modules/typeorm/browser/driver/react-native/ReactNativeDriver.js
+    // see: https://github.com/typeorm/typeorm/issues/2158
+    // this repo may not have this problem since using vite, but polkadot.js app will have
+    this.#chainPromise = setup({
+      endpoint: endpoint,
+      mockSignatureHost: true,
+      db: 'chopsticks.db',
```
xlc, line 59: this should come from provider props and should not have db cache enabled by default. for the use case of polkadot.js apps, it is unlikely the db cache going to be useful anyway

ermalkaleci:
the only way to make this work on polkadot.js.org/apps is to run chopsticks on background, otherwise this is unusable

xlc:
@ermalkaleci you mean the wasm execution will freeze UI by blocking the main therad? in that case yeah we need a web worker

ermalkaleci:
> @ermalkaleci you mean the wasm execution will freeze UI by blocking the main therad? in that case yeah we need a web worker

yes

ermalkaleci:
I think ChopsticksProvider can start a web worker that contains blockchain setup and use post message for rpc calls

packages/core/src/chopsticks-provider.ts
```diff
@@ -0,0 +1,287 @@
...
+  callback: ProviderInterfaceCallback
+  method: string
+  params: unknown[]
+  start: number
+  subscription?: SubscriptionHandler | undefined
+}
+
+export interface ChopsticksProviderProps {
+  /** upstream endpoint */
+  endpoint: string | undefined
```
xlc, line 35:
```suggestion
  endpoint: string
```

packages/core/src/chopsticks-provider.ts
```diff
@@ -0,0 +1,287 @@
...
+    subscription?: SubscriptionHandler,
+  ): Promise<T> => {
+    return new Promise<T>((resolve, reject): void => {
+      try {
+        if (!this.isConnected || this.#worker === undefined) {
+          throw new Error('Api is not connected')
+        }
+
+        if (method !== 'system_health') {
+          logger.info('[Chopsticks provider] send', { method, params })
```
xlc, line 148: why exclude `system_health`?
also could make this debug level

packages/core/src/chopsticks-provider.ts
```diff
@@ -0,0 +1,287 @@
...
+        }
+
+        const id = `${method}::${Date.now()}::${Math.random()}`
```
xlc, line 151: can we just make it `${method}::${this.idCounter++}`, also helps with debugging

packages/e2e/src/chopsticks-provider.test.ts
```diff
@@ -0,0 +1,99 @@
+import { ApiPromise } from '@polkadot/api'
+import { ChopsticksProvider } from '@acala-network/chopsticks-core'
+import { afterAll, beforeAll, describe, expect, it } from 'vitest'
+
+import { env, expectHex, expectJson, mockCallback, testingPairs } from './helper'
+
+// TODO: to be enabled after impl worker thread for nodejs compatibility
```
xlc, line 7: write a followup issue if you don't plan address this in this PR and put the issue number here so we don't forget to do it in future

packages/core/tsconfig.json
```diff
@@ -4,6 +4,7 @@
 		"outDir": "lib",
 		"rootDir": "src",
 		"target": "es2016",
+    "module": "esnext",
```
xlc, line 7: indent

packages/core/src/chopsticks-provider.ts
qiweiii: got too many logs from system health in development, forgot to recover

qiweiii:
Todo in future: 
node.js compatibility, issue: https://github.com/AcalaNetwork/chopsticks/issues/446

packages/core/src/chopsticks-worker.ts
```diff
@@ -0,0 +1,99 @@
...
+      })
+    }
+  },
+}
+
+onmessage = async (e) => {
+  switch (e.data.type) {
+    case 'connect':
+      try {
+        logger.debug('[Chopsticks worker] onMessage: connect. Initializing...')
```
xlc, line 47: no need to prefix message. should use child logger

`const logger = defaultLogger.child({ name: 'worker' })`

xlc (approved):

Issue #436, opened 2023-10-12
improve unsafeBlockHeight handling


xlc:
```
/Users/xiliangchen/projects/acala/chopsticks/packages/core/src/blockchain/block-builder.ts:170
    throw new Error('unsafeBlockHeight is not allowed to be less than current block number')
          ^
Error: unsafeBlockHeight is not allowed to be less than current block number
    at buildBlock (/Users/xiliangchen/projects/acala/chopsticks/packages/core/src/blockchain/block-builder.ts:170:11)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async TxPool.#buildBlock (/Users/xiliangchen/projects/acala/chopsticks/packages/core/src/blockchain/txpool.ts:232:43)
    at async TxPool.#buildBlockIfNeeded (/Users/xiliangchen/projects/acala/chopsticks/packages/core/src/blockchain/txpool.ts:212:7)
```

It should fail on the RPC but not crash the process

PR #447, opened 2023-10-17
fix unsafe block height error

qiweiii:
closes: https://github.com/AcalaNetwork/chopsticks/issues/436


xlc (commented):
Need a regression test that was failing and fixed by this

xlc (approved):

Issue #444, opened 2023-10-15
Invalid RPC request (Batch request)


Tbaut:
Chopsticks outputs an error for a batch request, as far as I understand. It seems that it may not support batch rpc calls?
Here's the Chopsticks error, the payload contains 2 calls:

```bash
[17:06:10.486] INFO (ws/66001): Invalid request: [{"method":"state_getStorage","params":["0x26aa394eea5630e07c48ae0c9558cef780d41e5e16056765bc8461851072c9d7","0xfd5bbc6e2cd2221a2f608b51c63ca5e1f124a92980f7ab6eafd581c49d2b6c27"],"id":7,"jsonrpc":"2.0"},{"method":"state_getStorage","params":["0x26aa394eea5630e07c48ae0c9558cef780d41e5e16056765bc8461851072c9d7","0xe786b54466eb089a6c96f7d6849ecb5d2fc21ff03bd9e301887d5bef72d0b200"],"id":8,"jsonrpc":"2.0"}]

```

While the flow of requests is the following, notice the last call (7 and 8) contains 2 `state_getStorage`.
```bash
17:06:10 DEBUG sqd:processor:rpc rpc call
                                 rpcUrl: ws://localhost:8000
                                 rpcId: 6
                                 rpcMethod: chain_getBlock
                                 rpcParams: ["0xfd5bbc6e2cd2221a2f608b51c63ca5e1f124a92980f7ab6eafd581c49d2b6c27"]
17:06:10 DEBUG sqd:processor:rpc rpc send
                                 rpcUrl: ws://localhost:8000
                                 rpcId: 6
17:06:10 DEBUG sqd:processor:rpc rpc response
                                 rpcUrl: ws://localhost:8000
                                 rpcId: 6
                                 rpcMethod: chain_getBlock
                                 rpcParams: ["0xfd5bbc6e2cd2221a2f608b51c63ca5e1f124a92980f7ab6eafd581c49d2b6c27"]
[...]

PR #453, opened 2023-10-18
support batch request

qiweiii:
closes: #444 

image

### DIFF
--- a/packages/chopsticks/src/server.ts
+++ b/packages/chopsticks/src/server.ts
@@ -6,12 +6,16 @@ import { defaultLogger, truncate } from './logger'
 
 const logger = defaultLogger.child({ name: 'ws' })
 
-const requestSchema = z.object({
+const singleRequest = z.object({
   id: z.number(),
   jsonrpc: z.literal('2.0'),
   method: z.string(),
   params: z.array(z.any()).default([]),
 })
+
+const batchRequest = z.array(singleRequest)
+
+const requestSchema = z.union([singleRequest, batchRequest])
 
 export type Handler = (
   data: { method: string; params: string[] },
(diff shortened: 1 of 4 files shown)


qiweiii:
Do we need to batch response?

xlc:
yes. it should do the same thing as a normal Substrate client

xlc:
and should have some tests

xlc (approved):

packages/chopsticks/src/server.ts
```diff
@@ -126,36 +164,18 @@ export const createServer = async (handler: Handler, port?: number) => {
...
-        send({
-          id: req.id,
-          jsonrpc: '2.0',
-          error: e instanceof ResponseError ? e : { code: -32603, message: `Internal ${e}` },
-        })
+      if (Array.isArray(req)) {
+        logger.trace({ req }, 'Received batch request')
+        const resp: Awaited<ReturnType<typeof processRequest>>[] = []
+        for (const r of req) {
+          const result = await processRequest(r)
```
ermalkaleci, line 171: Make this Promise.all([req.map(processRequset)])

packages/chopsticks/src/server.ts
```diff
@@ -126,36 +164,14 @@ export const createServer = async (handler: Handler, port?: number) => {
...
-      } catch (e) {
-        logger.info('Error handling request: %s %o', e, (e as Error).stack)
-        send({
-          id: req.id,
-          jsonrpc: '2.0',
-          error: e instanceof ResponseError ? e : { code: -32603, message: `Internal ${e}` },
-        })
+      if (Array.isArray(req)) {
+        logger.trace({ req }, 'Received batch request')
+        const resp = await Promise.all([...req.map(processRequest)])
```
ermalkaleci, line 169:
```suggestion
        const resp = await Promise.all(req.map(processRequest))
```
Small change

ermalkaleci (approved):

packages/testing/src/index.ts
```diff
@@ -81,10 +82,21 @@ export const setupContextWithConfig = async ({ timeout, ...config }: SetupConfig
...
+  })
+  await isWsClientReady
+
   return {
     url,
     chain,
+    /** @polkadot/api WsProvider */
     ws,
+    /** WebSocket Client */
+    wsClient,
```
ermalkaleci, line 99: do we need this?

qiweiii: not sure if there are other ways to test this in e2e test

packages/e2e/src/batch-request.test.ts
```diff
@@ -0,0 +1,48 @@
+import { afterAll, describe, expect, it } from 'vitest'
+
+import networks from './networks'
+
+describe('Batch request', async () => {
+  const acala = await networks.acala()
+  const { chain, wsClient } = acala
```
ermalkaleci, line 7:
```suggestion
  const { chain, url } = acala
  const webSocket = new WebSocket(url)
```

maybe we can do this?

qiweiii: yea, we can do this, but I feel it's cleaner to set it up in testing, and we may reuse it in future

ermalkaleci: if we keep then it needs a proper error handling. Current code will not resolve if connect fails. So better remove it

qiweiii: yes, we need to handle connect fail, but we still need to connect right? no matter where we put this code

ermalkaleci: yes, put it on the batch test

Issue #446, opened 2023-10-17
Node.js compatible chopsticks provider


qiweiii:
Now only support browser worker, need to support node.js as well

qiweiii:
No need this if wasm-executor worker PR works

PR #455, opened 2023-10-19
add chopsticks provider

qiweiii:
closes: #446 
- [x] e2e test works
- [x] web-test works
- [ ] copy build to polkadot.js app run locally, fix build errors
- [ ] test in polkadot.js app
- [ ] playwright test does not work (maybe do in future)


packages/core/src/chopsticks-provider.ts
```diff
@@ -1,286 +1,245 @@
...
+  get isConnected(): boolean {
+    return this.#isConnected
+  }
+
+  get isReady(): Promise<void> {
+    return this.#isReadyPromise
+  }
+
+  clone = (): ProviderInterface => {
+    return new ChopsticksProvider({ endpoint: this.#endpoint })
```
ermalkaleci, line 104: This doesn't feel right. Cloning provider will duplicate the chain. Maybe we can make provider to accept either endpoint or chainPromise so when cloning we create a provider with chainPromise

packages/core/src/chopsticks-provider.ts
```diff
@@ -1,286 +1,245 @@
...
+  #chain: Blockchain | undefined
+
+  constructor({ endpoint, blockHash, dbPath, storageValues }: ChopsticksProviderProps) {
+    if (!endpoint) {
+      throw new Error('ChopsticksProvider requires the upstream endpoint')
+    }
+    this.#endpoint = endpoint
+    this.#blockHash = blockHash
+    this.#dbPath = dbPath
+    this.#storageValues = storageValues
```
ermalkaleci, line 72: No need to duplicate stuff here. Just make chain public so setStorage can be called outside after provider is ready

packages/web-test/src/index.tsx
```diff
@@ -1,9 +1,43 @@
...
+	})
+	globalThis.api = api
+	api.isReady.then(() => {
+		api.rpc('new_block')
+		api.tx.balances.transfer(bob.address, 1000).signAndSend(alice, () => console.log('sent'))
+	})
+} catch (e) {
+	console.log(e)
+}
+
```
ermalkaleci, line 40: Also move this when we setupChain so we connect with same chain

PR #468, opened 2023-10-25
improve references

ermalkaleci:


xlc (commented):
some description and tests will be good

ermalkaleci:
> some description and tests will be good

we're removing blocks from `#blocksByNumber` but not from `#blocksByHash`, so block had still a strong reference. Also parent block was a strong reference keeping block in memory. This PR is addressing those issues

xlc (approved):

PR #476, opened 2023-10-27
run block rpc

xlc:
Replay a block and get all the details out from it

### DIFF
--- a/packages/chopsticks/src/plugins/run-block/index.ts
+++ b/packages/chopsticks/src/plugins/run-block/index.ts
@@ -37,8 +37,8 @@ export const cli = (y: yargs.Argv) => {
     async (argv) => {
       const context = await setupContext(argv as Config, true)
 
-      const header = await context.chain.head.header
-      const block = context.chain.head
+      const header = await context.parentBlock.header
+      const block = context.parentBlock
       const parent = await block.parentBlock
       if (!parent) throw Error('cant find parent block')
       const wasm = await parent.wasm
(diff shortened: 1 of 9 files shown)


qiweiii (approved):

packages/chopsticks/src/plugins/run-block/index.ts
```diff
@@ -82,3 +86,231 @@ export const cli = (y: yargs.Argv) => {
...
+  const { includeRaw, includeParsed, includeBlockDetails, head, block } = schema.parse(params)
+
+  const includeRawStorage = includeRaw ?? true
+
+  if (head) {
+    const headBlock = await chain.getBlock(head)
+    if (!headBlock) {
+      throw Error(`Invalid block hash ${head}`)
+    }
+    await chain.setHead(headBlock)
```
ermalkaleci, line 200: this doesn't seem right, why changing chain head?

packages/core/src/blockchain/block.ts
```diff
@@ -164,6 +165,17 @@ export class Block {
...
 
+  async read<T extends string>(type: T, query: StorageEntry, ...args: any[]) {
+    const key = compactHex(query(...args))
+    const value = await this.get(key)
+    if (!value) {
+      return undefined
+    }
+
+    const registry = await this.registry
+    return registry.createType(type, value)
```
ermalkaleci, line 176:
```suggestion
    return registry.createType(type, hexToU8a(value))
```

packages/chopsticks/src/plugins/run-block/index.ts
xlc: yeah no need to do it

PR #478, opened 2023-10-27
fix deleted key

ermalkaleci:


packages/e2e/src/decoder.test.ts
```diff
@@ -41,7 +40,7 @@ describe('decoder', async () => {
     const data = { data: { free: 10000000000 } }
     const value = meta.registry.createType('AccountInfo', data)
     expect(decodeKeyValue(meta, chain.head, SYSTEM_ACCOUNT, value.toHex())).toMatchSnapshot()
-
+    await new Promise((resolve) => setTimeout(resolve, 1000))
```
xlc, line 43: why the wait?

ermalkaleci: Nothing to do with the key fix. This can help with hanging test which I was getting randomly. Above test has a delay and I tried with this one and seems to work

xlc: can you comment the code to explain this? and can we reduce the delay amount to something like 100ms? 1s is long time

PR #489, opened 2023-10-30
dual publish

### LINKED ISSUE #487: Setup dual CJS/ESM packages
So that we can avoid this issue in downstream projects https://github.com/polkadot-js/api/issues/5636

https://nodejs.org/api/packages.html#dual-commonjses-module-packages

qiweiii:
close #487 

build for both cjs and esm.

- [x] test passed
- [x] web-test works
- [x] polkadot.js works


xlc (approved):

ermalkaleci:
Only core needs to be dual and this is not dual

qiweiii:
> Only core needs to be dual and this is not dual

@ermalkaleci could you explain a bit

ermalkaleci:
my understanding of dual is when you have `.cjs` and `.mjs` file extension and node will import `.cjs` and if you enable esm it will choose `.mjs` file extension

qiweiii:
> my understanding of dual is when you have `.cjs` and `.mjs` file extension and node will import `.cjs` and if you enable esm it will choose `.mjs` file extension

Yes, this is another way to do this

ermalkaleci:
I don't think we need this anyway. Bundlers will just transform the code and bun for example works with both.

xlc:
> another thing. Do not change current path. I'm sure it will break somewhere. Just add esm into `/lib/esm`. The tests are all running ok because our ts setup doesn't require any pre-build to work

assuming everything are wired correctly (the export map), everything should work

packages/chopsticks/tsconfig.esm.json
```diff
@@ -0,0 +1,8 @@
+{
+	"extends": "./tsconfig.json",
+	"compilerOptions": {
+		"module": "ES2020",
```
ermalkaleci, line 4: ES2021 because we use weakref
```suggestion
		"module": "ES2021",
```

ermalkaleci:
keep the current tsconfig and create new one only for esm so vscode won't give warnings

packages/chopsticks/tsconfig.esm.json
qiweiii: es2021 is not a valid module option, i will use es2022?

qiweiii:
> keep the current tsconfig and create new one only for esm so vscode won't give warnings

what warnings does it gove?

ermalkaleci:
tools will look for tsconfig.json which doesn't have any reference. we don't need to have 3 tsconfig. Just default one for cjs and another one for esm

ermalkaleci (approved):

PR #493, opened 2023-10-30
use parent block for wasm and metadata

xlc:
a block is executed with its parent's wasm. the only exception is genesis


xlc:
This is very unlikely going to make difference as the runtime upgrade blocks usually (100% of the time so far) doesn't include user tx and it really shouldn't break scheduled tasks without a migration

xlc:
we need a review of usage of wasm

Issue #190, opened 2023-02-08
follow mode


xlc:
always follow latest block on upstream
this means we need to disable various features such as block building and storage override

PR #498, opened 2023-11-01
follow mode

qiweiii:
close: https://github.com/AcalaNetwork/chopsticks/issues/190


packages/chopsticks/src/plugins/follow-chain/index.ts
```diff
@@ -0,0 +1,50 @@
...
+import { ProviderInterface } from '@polkadot/rpc-provider/types'
+import { defaultLogger } from '@acala-network/chopsticks-core'
+import _ from 'lodash'
+import type yargs from 'yargs'
+
+import { Config } from '../../schema'
+import { defaultOptions } from '../../cli-options'
+import { setupContext } from '../../context'
+
+const options = _.pick(defaultOptions, ['endpoint', 'wasm-override', 'runtime-log-level', 'offchain-worker'])
```
qiweiii, line 11: any missing options or redundant ones?

packages/chopsticks/src/plugins/follow-chain/index.ts
```diff
@@ -0,0 +1,50 @@
...
+const logger = defaultLogger.child({ name: 'follow-chain' })
+
+export const cli = (y: yargs.Argv) => {
+  y.command(
+    'follow-chain',
+    'Always follow the latest block on upstream',
+    (yargs) => yargs.options(options),
+    async (argv) => {
+      const { chain } = await setupContext(argv as Config, true)
```
qiweiii, line 20: do we want to setup a server instead?

qiweiii: also overrideParent is true, wasm-override will be applied to parent

packages/chopsticks/src/plugins/follow-chain/index.ts
```diff
@@ -0,0 +1,50 @@
...
+
+      let provider: ProviderInterface
+      const endpoint = argv.endpoint as string
+      if (/^(https|http):\/\//.test(endpoint || '')) {
+        provider = new HttpProvider(endpoint as string)
+      } else {
+        provider = new WsProvider(endpoint as string)
+      }
+      const apiPromise = await ApiPromise.create({
```
qiweiii, line 29: using an ApiPromise here to subscribe, is there a another way? like adding method in chain.api

packages/chopsticks/src/plugins/follow-chain/index.ts
```diff
@@ -0,0 +1,50 @@
...
+          SetEvmOrigin: {
+            extrinsic: {},
+            payload: {},
+          },
+        },
+      })
+
+      await apiPromise.isReady
+
+      apiPromise.rpc.chain.subscribeFinalizedHeads(async (header) => {
```
qiweiii, line 41: when get a block, is `setHead` enough?

packages/chopsticks/src/plugins/follow-chain/index.ts
xlc: yeah it is nice to have the server so we can check the follow status

packages/chopsticks/src/plugins/follow-chain/index.ts
```diff
@@ -0,0 +1,50 @@
...
+    'follow-chain',
+    'Always follow the latest block on upstream',
+    (yargs) => yargs.options(options),
+    async (argv) => {
+      const { chain } = await setupContext(argv as Config, true)
+
+      let provider: ProviderInterface
+      const endpoint = argv.endpoint as string
+      if (/^(https|http):\/\//.test(endpoint || '')) {
+        provider = new HttpProvider(endpoint as string)
```
xlc, line 25: subscribe is not supported by http provider so should throw an error here

packages/chopsticks/src/plugins/follow-chain/index.ts
xlc: another way is subscribe directly from the WsProvider

packages/chopsticks/src/plugins/follow-chain/index.ts
xlc: we want to run the block. something like this https://github.com/AcalaNetwork/chopsticks/blob/master/packages/chopsticks/src/plugins/run-block/index.ts#L53

and should have an option to choose to follow finalized heads or best heads

packages/chopsticks/src/plugins/follow-chain/index.ts
ermalkaleci: yeah try do avoid creating ApiPromise. It's better to create an api method that does the subscription

packages/core/src/api.ts
```diff
@@ -149,4 +150,18 @@ export class Api {
       return this.#provider.send<HexString[]>('state_getKeysPaged', params, !!hash)
     }
   }
+
+  async subscribeRemoteNewHeads(cb: ProviderInterfaceCallback) {
+    if (!(this.#provider instanceof WsProvider)) {
```
ermalkaleci, line 155: this.#provider.hasSubscriptions

packages/core/src/api.ts
```diff
@@ -149,4 +150,18 @@ export class Api {
...
+      throw new Error('subscribeRemoteNewHeads only works with WsProvider')
+    }
+    return this.#provider.subscribe('', 'chain_subscribeNewHeads', [], cb)
+  }
+
+  async subscribeRemoteFinalizedHeads(cb: ProviderInterfaceCallback) {
+    if (!(this.#provider instanceof WsProvider)) {
+      throw new Error('subscribeRemoteFinalizedHeads only works with WsProvider')
+    }
+    return this.#provider.subscribe('chain', 'chain_subscribeFinalizedHeads', [], cb)
```
ermalkaleci, line 165:
```suggestion
    return this.#provider.subscribe('chain_finalizedHead', 'chain_subscribeFinalizedHeads', [], cb)
```

packages/core/src/api.ts
```diff
@@ -149,4 +150,18 @@ export class Api {
       return this.#provider.send<HexString[]>('state_getKeysPaged', params, !!hash)
     }
   }
+
+  async subscribeRemoteNewHeads(cb: ProviderInterfaceCallback) {
+    if (!(this.#provider instanceof WsProvider)) {
+      throw new Error('subscribeRemoteNewHeads only works with WsProvider')
+    }
+    return this.#provider.subscribe('', 'chain_subscribeNewHeads', [], cb)
```
ermalkaleci, line 158:
```suggestion
    return this.#provider.subscribe('chain_newHead', 'chain_subscribeNewHeads', [], cb)
```

packages/chopsticks/src/plugins/follow-chain/index.ts
```diff
@@ -0,0 +1,93 @@
...
+            if (error) throw error
+            logger.info({ header: data }, `Follow ${argv.headMode} head from upstream`)
+            const parent = await chain.getBlock(data.parentHash)
+            if (!parent) throw Error(`Cannot find parent', ${data.parentHash}`)
+            const registry = await parent.registry
+            const header = registry.createType('Header', data)
+            const wasm = await parent.wasm
+
+            const block = await chain.getBlock(header.hash.toHex())
+            if (!block) throw Error(`Cannot find block ${header.hash.toHex()}`)
```
ermalkaleci, line 57:
```suggestion
            const header = registry.createType<Header>('Header', data)
            const wasm = await parent.wasm

            const block = new Block(chain, header.number.toNumber(), header.hash.toHex(), parent)
```

ermalkaleci (approved):

PR #352, opened 2023-07-29
browser support

ermalkaleci:

PR #505, opened 2023-11-01
fix: order of wasm and storage overrides

wischli:
In #352, the order of WASM and storage override order introduced in #199 was reverted. I suppose this happened by accident as the inline docs still state WASM override to have order priority. Moreover, it destroys the purpose of WASM override as new pallets cannot override storage.


ermalkaleci:
yes it's been reordered by accident

PR #511, opened 2023-11-04
fix esm

ermalkaleci:


packages/chopsticks/src/utils/generate-html-diff.ts
```diff
@@ -2,13 +2,13 @@ import { Block } from '@acala-network/chopsticks-core'
 import { HexString } from '@polkadot/util/types'
 import { decodeStorageDiff } from './decoder'
 import { mkdirSync, readFileSync, writeFileSync } from 'node:fs'
-import { template } from 'lodash'
+import _ from 'lodash'
 import url from 'node:url'
 
 export const generateHtmlDiff = async (block: Block, diff: [HexString, HexString | null][]) => {
   const { oldState, delta } = await decodeStorageDiff(block, diff)
   const htmlTemplate = readFileSync(url.resolve(__filename, './template/diff.html'), 'utf-8')
```
qiweiii, line 10: this `__filename` also need to fix, not available in esm, I think we can swich to use a bundler, so we can use these stuff in code

ermalkaleci: yes I know but it's used by cli so most likely it's fine. also I want to remove this completely in the future

PR #513, opened 2023-11-06
add execute block call in follow-chain

qiweiii:
closes: #70


qiweiii:
An error I get is 
```
 [11:05:18.948] ERROR (follow-chain/49419): Error when processing new head
    err: {
      "type": "Error",
      "message": "wasm `unreachable` instruction executed",
      "stack":
          Error: wasm `unreachable` instruction executed
              at Object.callback (/Users/qiweiyang/GitHub/chopsticks/packages/chopsticks/src/plugins/follow-chain/index.ts:134:23)
    }
```

Any clue how I should debug this? I can confirm the wasm I used is the one with try-runtime feature

xlc:
it should log more information if you have `LOG_LEVEL=trace`.

ermalkaleci:
what wasm are you using? looks like something wrong with wasm file

packages/chopsticks/src/plugins/follow-chain/index.ts
```diff
@@ -54,33 +61,80 @@ export const cli = (y: Argv) => {
             const header = registry.createType<Header>('Header', data)
             const wasm = await parent.wasm
 
-            const block = new Block(chain, header.number.toNumber(), header.hash.toHex(), parent)
+            const block = new Block(chain, header.number.toNumber(), header.hash.toHex(), parent, {
+              storage: parent.storage,
```
ermalkaleci, line 65: you can include header as well so it doesn't have to fetch again
```suggestion
            const block = new Block(chain, header.number.toNumber(), header.hash.toHex(), parent, {
              header,
              storage: parent.storage,
```

qiweiii:
> what wasm are you using? looks like something wrong with wasm file

I am using wasm build from this

https://github.com/AcalaNetwork/Acala/blob/7198bb16bc6078cd574a2545aa0306161e2d196a/Makefile#L138

ermalkaleci:
Remove true from setupContext

ermalkaleci:
this can't be implemented on follow-mode

Issue #512, opened 2023-11-04
Don't implicit load plugins


xlc:
Add a method to load plugins and call it from the entrypoint.
We don't want to load then on import the files to avoid impact of start time on other applications using chopsticks as a library.

ermalkaleci:
I think we worry about this when this becomes a problem

ermalkaleci:
can't we just remove the logs?

xlc:
Let's fix the problem, not the symptom

PR #514, opened 2023-11-06
improve load plugin

qiweiii:
related: #512

- only fix the rpc plugin load part
- for cli plugin, seems need to add another plugin command, so we can load cli plugin on-demand as sub-commands, but this will bring breaking change to how user use plugin cli


ermalkaleci:
if we do this then we need to a way to know the commands `-h` should tell you the commands also `rpc_methods` should return the methods

packages/chopsticks/src/plugins/index.ts
```diff
@@ -1,45 +1,51 @@
...
+export const rpcPluginHandlers: Handlers = {}
 
 // list of plugins directory
 const plugins = readdirSync(new URL('.', import.meta.url)).filter((file) =>
   lstatSync(new URL(file, import.meta.url)).isDirectory(),
 )
 
+// find all rpc exports
+export const rpcPluginMethods = readdirSync(new URL('.', import.meta.url))
+  .filter((file) => existsSync(new URL(`${file}/rpc.ts`, import.meta.url)))
```
ermalkaleci, line 19: this won't work when it's build

ermalkaleci: don't use exists. use lstatSync instead and see if there's a filename with rpc

executor/package.json
```diff
@@ -8,8 +8,8 @@
 		"url": "https://github.com/AcalaNetwork/chopsticks"
 	},
 	"scripts": {
-		"clean": "rm -rf web node",
-		"build": "wasm-pack build --target web --out-dir browser; wasm-pack build --target nodejs --out-dir node; scripts/pack-wasm.cjs"
+		"clean": "rm -rf browser node",
+		"build": "npx wasm-pack build --target web --out-dir browser && npx wasm-pack build --target nodejs --out-dir node && scripts/pack-wasm.cjs"
```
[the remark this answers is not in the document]
qiweiii: When work in dev container, it says wasm-pack not found

packages/chopsticks/src/plugins/index.ts
qiweiii: both `exist` and `lstatSync` won't work, they both require full filename with extension, so I changed to filter by file name instead

executor/package.json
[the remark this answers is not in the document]
ermalkaleci: yarn install should install it

PR #520, opened 2023-11-07
use rollup

qiweiii:
Got errors saying `Module not found` when trying to update polkadot.js app PR with latest version.
Tried a few fix with swc setup, but did not make it work.

Then I checked a few packages with dual bundle, I tried to just follow vite-node’s solution: https://github.com/vitest-dev/vitest/blob/main/packages/vite-node/rollup.config.js, it seems all work fine

Tests
  - [x]  run cli
  - [x]  e2e
  - [x]  web-test
  - [x]  playwright-test
  - [x]  polkadot.js app
  - [x]  mangata-e2e


ermalkaleci:
webpack thing
`{
            test: /\.m?js$/,
            resolve: {
              fullySpecified: false, // disable the behaviour
            },
          },`

qiweiii:
don't need this for now

Issue #506, opened 2023-11-01
cap cache size


xlc:
https://github.com/AcalaNetwork/chopsticks/blob/40d9e505a9171c027b2fe55868ac8b447bab70fe/packages/core/src/utils/decoder.ts#L11-L18

Need to cap the cache size

PR #530, opened 2023-11-09
max cache for decoder

qiweiii:
closes: #506

a simple lru-cache for decoder cache


packages/core/src/utils/decoder.ts
```diff
@@ -2,26 +2,35 @@ import '@polkadot/types-codec'
...
 
 import { decodeWellKnownKey } from './well-known-keys.js'
 
-const _CACHE: Record<string, Map<HexString, StorageEntry>> = {}
+const _CACHE: Record<string, LRUCache<HexString, StorageEntry>> = {}
 
-const getCache = (uid: string): Map<HexString, StorageEntry> => {
+function createCache() {
+  return new LRUCache<HexString, StorageEntry>({
+    max: 500, // The maximum number of items in the cache
```
xlc, line 17: can we do some quick tests to see if this number is good?
basically make max to a very big number, run bunch blocks, check cache size and memory usage

qiweiii: I tested with some polkadot and acala blocks, the number of items in this cache rarely go up to more than 50, so I changed max to 50. but memory usage keeps growing, there are other places that is eating the memory (assuming lru-cache is working fine)

xlc (approved):

Issue #534, opened 2023-11-10
new package chopsticks-utils


xlc:
move all the code thats not related to `vitest` from testing to utils. I want to write some quick Chopsticks script ([example](https://github.com/xlc/chopsticks-ksm-xcm/tree/master)) without importing vitest

ermalkaleci:
it's just a file. why not move it to chopsticks?

xlc:
we will have more things there in future

qiweiii:
Do we only want test related utils in this package? is so we could name it testing-utils

xlc:
For example, to implement this https://github.com/polkadot-fellows/runtimes/issues/83, we need the ability to launch new networks on the fly to dry run the xcm on dest chain. This will be something can be implemented here and not just for testing.

PR #536, opened 2023-11-10
add utils package

qiweiii:
closes: #534


ermalkaleci:
This is a package with just one file. I will prefer to just move to chopsticks

xlc (approved):

Issue #515, opened 2023-11-06
storage override needs to support storage value type of `()` 


xlc:
for example, `whitelist.whitelistdCall`

qiweiii:
how do I reproduce this?

xlc:
<img width="790" alt="Screenshot 2023-11-09 at 8 55 09 PM" src="https://github.com/AcalaNetwork/chopsticks/assets/1201310/6b46ca50-98d1-48da-91ea-327a475abb89">

this is on Polkadot. 

We need a way to use `dev_setStorage` to add and remove whitelistedCall entry. The value type is `()`, which is basically a `void` so we may need to distinguish null and undefined

ermalkaleci:
@xlc just pass '0x' for empty values

```
Whitelist:
    WhitelistedCall:
      -
        -
          - '0xe284be84dcfaf714ef2b7717b54914632406f2c17d8203d3268e4c4ca68fa144'
        - 0x
```

qiweiii:
Maybe we should add it in readme or in docs abt how to use these, so user knows

PR #542, opened 2023-11-13
storage null value not handled properly

qiweiii:
fixes: #515

### DIFF
--- a/packages/core/src/blockchain/storage-layer.ts
+++ b/packages/core/src/blockchain/storage-layer.ts
@@ -21,7 +21,7 @@ export interface StorageLayerProvider {
   /**
    * Get the value of a storage key.
    */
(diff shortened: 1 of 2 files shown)


xlc (approved):

ermalkaleci:
This is very tricky

qiweiii:
the storage-migrate test is not passed because before this pr, null value storage is marked as Deleted, so the storage-migrate test works, but now it allows null value...  so maybe there is some info that can help us distinguish different kinds of null values?

ermalkaleci:
leave this to me

ermalkaleci:
I think it should work. but let me confirm

ermalkaleci:
yes it works fine. Just value needs to be '0x'

ermalkaleci:
no change is needed

PR #543, opened 2023-11-13
bun missing URL.createObjectURL workaround for wasm init

ermalkaleci:


executor/scripts/pack-wasm.cjs
```diff
@@ -19,9 +19,9 @@ const BYTES = '${base64}';
 import { base64Decode, unzlibSync } from '@polkadot/wasm-util';
 const WASM_BYTES = unzlibSync(base64Decode(BYTES, new Uint8Array(LEN_IN)), new Uint8Array(LEN_OUT));
 
-import wasmInit from "./chopsticks_executor.js";
+import wasmInit, { initSync } from "./chopsticks_executor.js";
 const blob = new Blob([WASM_BYTES], { type: "application/wasm" });
-export const wasmReady = wasmInit(URL.createObjectURL(blob));
+export const wasmReady = typeof Bun !== undefined ? Promise.resolve(initSync(WASM_BYTES)) : wasmInit(URL.createObjectURL(blob));
```
xlc, line 24: best to add a comment explain the workaround and link to upstream issue so in future someone else can look at this code understand the workaround and check if we can remove the workaround

ermalkaleci: actually I will just use initSync. The reason I was using URL was because chrome will throw error if wasm isn't loaded async but now wasm is loaded on webworker. I tested and there's no warning.

ermalkaleci: #544

PR #555, opened 2023-11-23
support multiple endpoints

ermalkaleci:


packages/core/src/setup.ts
```diff
@@ -37,7 +37,7 @@ export const setup = async (options: SetupOptions) => {
   let provider: ProviderInterface
   if (options.genesis) {
     provider = options.genesis
-  } else if (/^(https|http):\/\//.test(options.endpoint || '')) {
+  } else if (typeof options.endpoint === 'string' && /^(https|http):\/\//.test(options.endpoint || '')) {
```
xlc, line 40: we should have a helper somehwere to create the provider

PR #566, opened 2023-11-28
add signer

xlc:


ermalkaleci:
@xlc we need a better way for this snapshot file. There's no way to tell what changed

xlc:
it is mostly because I am changing the block. Otherwise the diff will be very reasonable.

Issue #561, opened 2023-11-24
Add possibility of proving additional relay storage keys in set validation data


girazoki:
Chains like Tanssi require collators to prove additional relay storage keys (in this case, the para head of a particular parachain-id). The way chopsticks works right now is that in the `set_validation_data` a new` relay storage root` is computed and injected based, among other things, on the incoming `xcms` and the `relay slot number`.

It could be good that we can add a set of additional `key_values` that we can inject in the generation of the proof, in a similar manner to how we did it with rust: https://github.com/paritytech/polkadot-sdk/pull/1757. Tipically the relay storage root provided in `set_validation_data` is used to verify any other inherents proving relay-storage, so we need would need to take any additional key-value pair into account in the mocked version.

Additionally, it would be good if the trie-nodes generated by the proof were accessible for other inherents. Currently each inherent is quite isolated, and know from each other. Potentially there could be a way in which both `set_validation_data` and any other inherent use the same proofGenerator instance.

ermalkaleci:
can you give an example how you intend to use it?

girazoki:
yes we have our own sort of hacky implementation here: https://github.com/moondance-labs/chopsticks/pull/6

Essentially we generate a proof for additional keys in `set_validation_data`, and then we use these trie nodes in a the new inherent where the new keys are proved

ermalkaleci:
@girazoki is storage `authorInherent.highestSlotSeen()` best way to get slot from your chain?

girazoki:
sorry @ermalkaleci I somehow missed this! yes it is!

PR #569, opened 2023-11-28
refactor decode proof to extract all keys

ermalkaleci:


ermalkaleci:
this will make it easy to implement new inherent #561
decode_proof will find all possible keys so there's no need to inject unchanged ones

girazoki:
thank you!

girazoki:
@ermalkaleci as I am understanding it, now every key that was proven from the previous code and that is not modified will stay be included and proved on the new built block?

ermalkaleci:
yes, you don't need to touch `SetValidationData` inherent. SetValidationData will be executed first and set new proofs which you can read on your inherent from `query.parachainSystem.relayStateProof()`

girazoki:
ah that's amazing. I think it does cover most of our needs so we might propose a PR adding tanssi support. The only thing I am thinking is: what happens if there are some keys being proved in one inherent (set-validation-data) and others in another inherent (authorities-noting in our case) in production? Probably the nibbles passed to set-validation-data do not include all keys to be able to fork correctly.

We might propose the possibility to add additional keys to set-validation-data inherent if that's fine in the PR that supports tanssi (in other words, being able to add additional keys to the `WellKnownKeys` struct)

girazoki:
the other thing we can do is refactor our node to inject a single proof (in set-validation-data) that contains all keys we need to prove. That should work with this implementation and I think its cleaner!

ermalkaleci:
I think I was wrong. I will look again tomorrow, maybe it needs a refactor

PR #591, opened 2023-12-12
Add zeitgeist config yaml

Chralt98:


configs/zeitgeist.yml
Chralt98, line 1: Should I add `block: ${env.ZEITGEIST_BLOCK_NUMBER}`? How can the `ZEITGEIST_BLOCK_NUMBER` be supplied?

ermalkaleci: Environment Variables

ermalkaleci: you don't need config to fork a chain, you can fork any chain with just providing an endpoint. config just makes it easier because you have all the option defined into one file and users can fork your chain just by `-c acala` and it will load configs from gh repo `configs/acala.yml`. There are cases when you need to register custom types for chopsticks to work or you want to give alice some tokens. You can make config however it's best for the users

ermalkaleci: any config parameter can be overridden by option. e.g: `-c acala --endpoint=<new_endpoint> --block=<specific block>`

xlc: best to add it for consistency
the main use case is that you can have a .env file with it defined to a fixed number, and then rerun will be able to leverage the cache to make everything faster

configs/zeitgeist.yml
```diff
@@ -0,0 +1,29 @@
+endpoint: wss://main.rpc.zeitgeist.pm/ws
+# endpoint: wss://zeitgeist-rpc.dwellir.com
+# endpoint: wss://zeitgeist.api.onfinality.io/public-ws
+mock-signature-host: true
```
Chralt98, line 4:
```suggestion
mock-signature-host: true
block: ${env.ZEITGEIST_BLOCK_NUMBER}
```

configs/zeitgeist.yml
```diff
@@ -0,0 +1,30 @@
+endpoint: wss://main.rpc.zeitgeist.pm/ws
+# endpoint: wss://zeitgeist-rpc.dwellir.com
```
ermalkaleci, line 2: can you remove these or make it array. chopsticks supports multiple endpoints

configs/zeitgeist.yml
```diff
@@ -0,0 +1,30 @@
...
+# endpoint: wss://zeitgeist.api.onfinality.io/public-ws
```
Chralt98, line 3:
```suggestion
endpoint: 
  - wss://main.rpc.zeitgeist.pm/ws
  - wss://zeitgeist-rpc.dwellir.com
  - wss://zeitgeist.api.onfinality.io/public-ws
```

configs/zeitgeist.yml
Chralt98: https://github.com/AcalaNetwork/chopsticks/pull/591/commits/c0e51254d0723eaa61ee0d130c580bcd1b72cc58

ermalkaleci (approved):

PR #592, opened 2023-12-13
fix decode cache

xlc:
we need individual cache for each runtime version


packages/core/src/utils/decoder.ts
```diff
@@ -28,8 +28,8 @@ const getCache = (uid: string): LRUCache<HexString, StorageEntry> => {
   return _CACHE[uid]
 }
 
-const getStorageEntry = (meta: DecoratedMeta, block: Block, key: HexString) => {
-  const cache = getCache(block.chain.uid)
+const getStorageEntry = (meta: DecoratedMeta, key: HexString) => {
+  const cache = getCache(meta.registry.metadata.hash.toHex())
```
xlc, line 32: not sure if this will be slow, but that's to be addressed later when we do proper benchmarking

PR #606, opened 2023-12-24
fix storage import encode

ermalkaleci:


laurogripa:
Nice 👏 

You beat me to it, I started investigating but had to stop for xmas celebrations 😆 
Thanks for addressing the issue 👊

PR #640, opened 2024-01-23
add manta config

zqhxuyuan:


ermalkaleci:
your endpoint doesn't work. it's extremely slow and will timeout

configs/manta.yml
```diff
@@ -0,0 +1,20 @@
...
+   Key: dfbMFtQXqiKHH4d2y4CFcVQCpopqKcaEiYv2DCUpVCphoz8kB
+  System:
+    Account:
+      -
+        -
+          - dfbMFtQXqiKHH4d2y4CFcVQCpopqKcaEiYv2DCUpVCphoz8kB
+        - data:
+            free: 1000000000000000
+  ParachainStaking:
+    # NOTE: MANTA_BLOCK_NUMBER should set to 2170792 in .env file.
```
ermalkaleci, line 17: This config is not useful  if users have to manually change it. Configs in this repo are meant to start the fork without any aditional work, just `chopsticks -c acala` and it use the predefined config. This config should work for any block otherwise I think it's best to guide your users how to do a proper setup

ermalkaleci:
> your endpoint doesn't work. it's extremely slow and will timeout

now its a bit better

configs/manta.yml
ermalkaleci: see if you can find a way to cheat nimbus

xlc: if this config only works for a block number, then we should just hardcoded the block number here

ermalkaleci: I tried but it will build a few blocks then fail when nimbus rotates author

ermalkaleci:
#648

PR #646, opened 2024-01-25
update logging

ermalkaleci:


packages/core/src/logger.ts
```diff
@@ -1,14 +1,16 @@
...
   transport: {
     target: 'pino-pretty',
+    options: {
+      ignore: 'pid,hostname',
+      hideObject: !!process.env.LOG_COMPACT,
+    },
   },
 })
 
-export const defaultLogger = pinoLogger.child({ app: 'chopsticks' })
```
xlc, line 10: no I want this. otherwise I can't distinguish the logs from chopsticks library vs logs from other code in supx

Issue #604, opened 2023-12-22
tests to ensure cli parsing code is correct


xlc:
so we can catch bug like https://github.com/AcalaNetwork/chopsticks/pull/603

qiweiii:
I saw that we already have an [options.test.ts](https://github.com/AcalaNetwork/chopsticks/blob/d7b476eb44f14e8e88e39b2f263d9ccffe330c4f/packages/chopsticks/src/schema/options.test.ts), do we need more tests?

xlc:
it doesn't catch the bug I introduced here https://github.com/AcalaNetwork/chopsticks/pull/636 so we need more

PR #651, opened 2024-01-29
tests: add test for options parsing

qiweiii:
closes: #604


packages/chopsticks/src/schema/parse.test.ts
```diff
@@ -0,0 +1,60 @@
...
+
+describe('Parsed options', () => {
+  const defaults = {
+    port: 8000,
+    'build-block-mode': 'Batch',
+  }
+  it('parsed multi type options should work', () => {
+    expect(
+      configSchema.parse({
+        block: 4500000,
```
xlc, line 37: add a test for `block: '4500000'`

packages/chopsticks/src/schema/parse.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  return arrayOfFiles
+}
+
+describe('Existing configs', async () => {
+  const configPaths = getAllFiles(path.join(__dirname, '../../../../configs'))
+  it.each(configPaths.map((p) => ({ name: _.last(p.split('/')), path: p })))(
+    '$name config parsing is correct',
+    async ({ path }) => {
+      const config = await fetchConfig(path)
+      expect(configSchema.parse(config)).toMatchSnapshot()
```
xlc, line 24: this is useful, with one issue that add or modify config will need to regenerate snapshots. maybe the easy way is just not check the output but only check parse is successful

xlc:
I would like to test code in here as well https://github.com/AcalaNetwork/chopsticks/blob/3a8bcc792476c36448a3b28d3b8964f635bf80db/packages/core/src/setup.ts#L85-L104

qiweiii:
> I would like to test code in here as well
> 
> https://github.com/AcalaNetwork/chopsticks/blob/3a8bcc792476c36448a3b28d3b8964f635bf80db/packages/core/src/setup.ts#L85-L104

agree, probably I shoud add an e2e test that read console output

xlc (approved):

xlc (approved):

PR #653, opened 2024-01-29
add integritee and encointer configs

brenzi:
tested all and working

such a nice tool :+1:

### DIFF
--- a/configs/encointer-kusama.yml
+++ b/configs/encointer-kusama.yml
@@ -1,0 +1,20 @@
+endpoint:
+  - wss://kusama.api.encointer.org
+mock-signature-host: true
+block: ${env.ENCOINTER_KUSAMA_BLOCK_NUMBER}
+db: ./encointer-kusama.db.sqlite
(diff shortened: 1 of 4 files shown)


configs/encointer-kusama.yml
```diff
@@ -0,0 +1,25 @@
+# using chopsticks with encointer will not allow to perform actions involving encointer custom types which are
+# not supported by chopsticks: CommunityIdentifier, fixpoint balances
+# also, encointer uses custom rpc methods which are not available on chopsticks
+# still, basic chain functionality can be tested
```
brenzi, line 4: FYI.

xlc (approved):

configs/encointer-kusama.yml
ermalkaleci: Chopsticks implements only the basic PRCs. Your custom RPCs methods are part of the node. Chopsticks only uses your runtime. Regarding types, you can define custom types on your config file

PR #662, opened 2024-01-30
bump 0.9.8

ermalkaleci:


ermalkaleci:
@albertov19 new version is out

PR #673, opened 2024-02-22
acala evm tracing

ermalkaleci:
https://github.com/AcalaNetwork/Acala/pull/2716
`yarn start -c acala trace-transaction <tx-hash> --output trace.json`


packages/chopsticks/src/plugins/trace-transaction/index.ts
```diff
@@ -0,0 +1,168 @@
...
+      const txIndex = extrinsics.findIndex((tx) => blake2AsHex(tx) === txHash)
+
+      const newBlock = new Block(context.chain, block.number, blockHash, parent, {
+        header,
+        extrinsics: [],
+        storage: parent.storage,
+      })
+
+      meta.registry.register({
+        Step: {
```
xlc, line 80: the metadata from the overridden runtime should have the types?

ermalkaleci: for some reason it's not generating types, maybe because it's not used on extrinsic function. not sure yet

xlc: oh right. we need pjs to support metadata v15 for that to work

packages/chopsticks/src/plugins/trace-transaction/index.ts
```diff
@@ -0,0 +1,191 @@
...
+        u8aToHex(meta.registry.createType('u64', gasLimit).toU8a()),
+        u8aToHex(meta.registry.createType('u32', storageLimit).toU8a()),
+        '0x00', // empty access list
+      ])
+
+      const logs = meta.registry
+        .createType<any>(`Result<Vec<${config.vm ? 'Step' : 'CallTrace'}>, DispatchError>`, res.result)
+        .asOk.toJSON()
+
+      const filepath = `${process.cwd()}/trace-${txHash}.json`
```
xlc, line 185: generate random file is not good. should take path from user input.
the proper convention for cli tool is that the output are default to stdout, and logs are to stderr. in that way we can just `cmd > out.json`

xlc (approved):

Issue #630, opened 2024-01-11
Failing `dev_setStorage`


NachoPal:
After this commit: https://github.com/AcalaNetwork/chopsticks/commit/5c81c80e5c60bec090376cb53c58b2fa9f7570db
`dev_setStorage` stopped working for me.

I have:
```typescript
        await api.rpc('dev_setStorage', {
            preimage: {
                preimageFor: [
                    [
                        [[hashedCall, callLength]],
                        call
                    ]
                ]
            }
        })
```
I do not know if I have to do some modification to the arguments format.

ermalkaleci:
can you post more details, keys, values? I can't tell what you need to modify

NachoPal:
These are the logs:
```
[12:51:49.792] TRACE (ws/17807): Received single request
    app: "chopsticks"
    req: {
      "id": 13,
      "jsonrpc": "2.0",
      "method": "dev_setStorage",
      "params": [
        {
          "preimage": {
            "preimageFor": [
              [
                [
                  [
                    "0x4677e43381a888a45e7c72497a2e0c93f39d1805b72f70b14e32fa35a48af8b4",
                    11
                  ]
                ],
                "0x6304000100a10f03000000"
              ]
            ]
          }
        }
      ]
    }
[12:51:49.792] TRACE (ws/17807): Received message
    app: "chopsticks"
    id: 13
    method: "dev_setStorage"
[12:51:49.792] TRACE (17807): Handling dev_setStorage
    app: "chopsticks"
    module: "rpc"
[12:51:49.793] DEBUG (plugin/17807): Registered plugin set-storage RPC
    app: "chopsticks"
[12:51:49.794] DEBUG (17807): dev_setStorage
    app: "chopsticks"
    hash: "0x65f3279eb676bb6c73f67575e7477cbd6d78654b8df04d4ebadf785837304ec0"
    values: {
      "preimage": {
        "preimageFor": [
          [
            [
              [
                "0x4677e43381a888a45e7c72497a2e0c93f39d1805b72f70b14e32fa35a48af8b4",
                11
              ]
            ],
            "0x6304000100a10f03000000"
          ]
        ]
      }
    }
[12:51:49.794] TRACE (ws/17807): Response for request
    app: "chopsticks"
    id: 13
    method: "dev_setStorage"
    result: "0x65f3279eb676bb6c73f67575e7477cbd6d78654b8df04d4ebadf785837304ec0"
```

[...]

PR #675, opened 2024-02-29
Fix `--preimage` storage value

serban300:
Related to: https://github.com/AcalaNetwork/chopsticks/issues/630 and https://github.com/paritytech/polkadot-sdk/issues/3522

I'm trying to simulate some bridge-initialization calls using chopsticks. I used the code found in `dry-run-preimage.ts` as an example for crafting my own `dev_setStorage` call for simulating an approved preimage. The initial version didn't work (it was failing with `CallUnavailable`) and it took quite a while to find the solution presented in #630. Not sure if this is the best fix, but I think it would bring value to have this fixed. At least it would have saved a lot of time for me.


ermalkaleci:
dry-run-preimage works fine. what are you trying to fix?

serban300:
Oh, I see. Sorry, I didn't realize. Ok. I'll close the PR then.

PR #685, opened 2024-03-06
fix header rpc response

ermalkaleci:


xlc:
need some tests

ermalkaleci:
yes we can have some tests. it's very small change on method `processHeader`. Nothing can go wrong

xlc:
but the fact it wasn't exactly correct at the first place is because we don't have the right tests. we should always have regression tests

PR #689, opened 2024-03-11
connectHorizontal open any channel

ermalkaleci:


packages/core/src/xcm/horizontal.ts
```diff
@@ -30,5 +30,15 @@ export const connectHorizontal = async (parachains: Record<number, Blockchain>)
         }
       }
     })
+
+    const hrmpHeads = await chain.head.read('BTreeMap<u32, H256>', meta.query.parachainSystem.lastHrmpMqcHeads)
+    if (hrmpHeads && !process.env.DISABLE_AUTO_HRMP) {
```
xlc, line 35: best to avoid env in the core package. they are global variables so all the reason to avoid global variables applies.
we should just pass arguments into here, and should be the cli package to handle envs. also need to document them

ermalkaleci: Yes we have a few places when we use env, we need to address them

PR #695, opened 2024-03-15
feat: add grandpa justifications mock rpc

qiweiii:
add
- grandpa_subscribeJustifications
- grandpa_unsubscribeJustifications
- state_getReadProof


qiweiii:
is this impl what we expect? @xlc

qiweiii:
weired, cannot pass test, the callback is only called once..

ermalkaleci:
I'm afraid there's no use of fake data for these methods. These are required for beefy and needs to be  real data

xlc:
It is not possible to have real data. So the user of those RPCs (i.e. the bridge relayer) needs to have a special mode to skip signature validation, which is the plan.

packages/core/src/rpc/substrate/state.ts
```diff
@@ -56,6 +56,24 @@ export const state_getKeysPaged: Handler<[string, number, string, HexString], st
...
+ * @return mock proof
+ */
+export const state_getReadProof: Handler<
+  [HexString[], HexString],
+  { at: HexString; proof: HexString[] } | null
+> = async (_context, [keys, hash]) => {
+  return {
+    at: hash,
+    proof: keys.map(() => ('0x' + '7'.repeat(64)) as HexString),
+  }
```
serban300, line 74: Is there any way we could get a real proof here ? As far as I know, for this we don't need signing keys.

ermalkaleci: Yes

ermalkaleci: @serban300 I've implemented `state_getReadProof` to return proof of requested keys but the `trie_state_root` won't be the same as block `trie_state_root`. I don't know if correct hash is required and not sure if we can do that. Can you give this a try first before I look deeper into the implementation?

serban300: @ermalkaleci thanks ! Yes, will try this. Just have to prioritize some reviews first, and then I will get to this. Hopefully this week.

packages/core/src/rpc/substrate/grandpa.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const grandpa_subscribeJustifications: Handler<void, string> = async (context, _params, { subscribe }) => {
+  let update = (_block: Block) => {}
+
+  const id = context.chain.headState.subscribeHead((block) => update(block))
+  const callback = subscribe('grandpa_justifications', id, () => context.chain.headState.unsubscribeHead(id))
+
+  update = async (block: Block) => {
+    // Proof needs to be different for each block
+    callback(blake2AsHex(block.hash))
```
serban300, line 14: I'm getting an error when receiving such a proof:
```
[Polkadot_to_BridgeHubKusama_Sync] 2024-03-21 20:22:21 +00 ERROR bridge Failed to read justification target from the Polkadot justifications stream: "decode failed with error Error { cause: Some(Error { cause: Some(Error { cause: None, desc: \"Not enough data to fill buffer\" }), desc: \"Could not decode `Commit::target_hash`\" }), desc: \"Could not decode `GrandpaJustification::commit`\" }"
```

The proof needs to be an encoded `GrandpaJustification` that contains at least the `targetHash` and `targetNumber`. Something like:

```
    const justification: GrandpaJustification = {
        round: 0,
        commit: {
            targetHash: block.hash,
            targetNumber: block.number,
            precommits: []
        },
        votesAncestries: []
    };
```
And then it needs to be encoded. 

Not sure how to instantiate such a type here. Can you take a look please ?

ermalkaleci: @serban300 can you try latest commit?

serban300: @ermalkaleci just some context first. I haven't tested it yet, but I don't think this will work. Looks like it returns a beefy justification from `grandpa_subscribeJustifications` if I understand correctly. These should be 2 different APIs: [`grandpa_subscribeJustifications`](https://polkadot.js.org/docs/substrate/rpc/#subscribejustifications-justificationnotification) and [`beefy_subscribeJustifications`](https://polkadot.js.org/docs/substrate/rpc#subscribejustifications-beefyversionedfinalityproof). Also I would define them in 2 separate files. 

Another think I should mention is that we don't use the beefy justifications yet for the Polkadot<>Kusama bridge. We have a beefy light client, but it's still a work in progress, we don't use it now and it's not a priority. We only use grandpa. So we don't need `beefy_subscribeJustifications` yet. I don't know about the team working at the Ethereum bridge. They use beefy, but I don't know if they plan to perform tests using chopsticks.

For `grandpa_subscribeJustifications` we need to return an encoded `GrandpaJustification` struct as mentioned in the previous comment. For the moment I think a good enough test would be just to decode it and see that the `targetHash` and `targetNumber` match the latest finalized block if that's possible.

ermalkaleci: ok my bad. I was looking for justification on wested and I just found BEEFY. updating now

ermalkaleci: @serban300 updated

serban300: @ermalkaleci Thanks ! Now I'm not getting any decoding error anymore. And if I comment the finality proof verification logic in the relayer and in the bridge hubs runtimes, updating the best header for the bridged relay chain works.

packages/core/src/rpc/substrate/state.ts
serban300: Checked. This doesn't work. I'm getting an error when validating the storage proof. We would need the `trie_state_root` to be the same as the block `trie_state_root`. Since I think that's how we're validating the proof. Would that be possible ?

packages/core/src/rpc/substrate/grandpa.ts
ermalkaleci: Can you send me a link to the code you're disabling?

serban300: Sure. 

This is the disabled runtime code: https://github.com/serban300/polkadot-sdk/commit/06dcf0a4531adda2bd12d7b7b29974b244084dc4

And this is the disabled relayer code: https://github.com/serban300/parity-bridges-common/commit/f334bddbe4c1630a8957dae479af78c732f15649

packages/core/src/rpc/substrate/state.ts
ermalkaleci: I'm not sure

PR #700, opened 2024-03-21
process queued messages

ermalkaleci:


packages/chopsticks/src/schema/index.ts
```diff
@@ -59,6 +59,11 @@ export const configSchema = z.object({
         'Resume from the specified block hash or block number in db. If true, it will resume from the latest block in db. Note this will override the block option',
     })
     .optional(),
+  'process-queued-messages': z
+    .boolean({
+      description: 'Process queued messages when building block. Default to true. Set to false to disable it.',
```
xlc, line 64: Maybe something like: Produce extra block when queued messages are detected.
So that people know what exactly does the "process" means

packages/core/src/blockchain/txpool.ts
```diff
@@ -195,6 +196,25 @@ export class TxPool {
...
+      const prefix = '0xb8753e9383841da95f7b8871e5de326954e062a2cf8df68178ee2e5dbdf00bff'
+      const meta = await this.#chain.head.meta
+      const keys = await this.#chain.head.getKeysPaged({ prefix, pageSize: 1000 })
+      for (const key of keys) {
+        const rawValue = await this.#chain.head.get(key)
+        if (!rawValue) continue
+        const message = meta.registry.createType('PalletMessageQueueBookState', hexToU8a(rawValue)).toJSON() as any
+        if (message.size > 0) {
+          // build a new block to process the queued messages
+          await this.#chain.newBlock()
```
xlc, line 214: add a debug log here to indicate why another new block

xlc (approved):

PR #715, opened 2024-03-29
fix: process env usage with doc

### LINKED ISSUE #690: review usage of process.env
We should avoid using process.env in core package, and only use it in the cli. There should be a single file contains all the available envs and we need to document them.

qiweiii:
fix #690


packages/core/src/env.ts
```diff
@@ -0,0 +1,33 @@
...
+   * Disable plugins for faster startup. Default is `false`.
+   */
+  DISABLE_PLUGINS: z.boolean().default(false),
+  HTTP_PROXY: z.string().optional(),
+  HTTPS_PROXY: z.string().optional(),
+  /**
+   * Chopsticks log level, "fatal" | "error" | "warn" | "info" | "debug" | "trace".
+   * Default is "info".
+   */
+  LOG_LEVEL: z.string().default('info'),
```
ermalkaleci, line 19: can you add available options as well?

packages/core/src/env.ts
```diff
@@ -0,0 +1,33 @@
+import * as z from 'zod'
+
+const environmentSchema = z.object({
+  /**
+   * Disable auto HRMP on setup. Default is `false`.
+   */
+  DISABLE_AUTO_HRMP: z.boolean().default(false),
```
ermalkaleci, line 7: does it convert 'false' into boolean? can you add a test?

packages/chopsticks/src/utils/tunnel.ts
```diff
@@ -1,13 +1,14 @@
...
 bootstrap()
+import { environment } from '@acala-network/chopsticks-core'
 import npmConf from '@pnpm/npm-conf'
 
 const npmConfig = npmConf().config
 
 global.GLOBAL_AGENT.HTTP_PROXY =
-  process.env.HTTP_PROXY ||
+  environment.HTTP_PROXY ||
   process.env.http_proxy ||
```
xlc, line 10: should handle `http_proxy` and `https_proxy` as well

xlc (approved):

ermalkaleci (approved):

PR #749, opened 2024-05-04
fix: runtime call not found

### LINKED ISSUE #748: handle runtime call not found error
https://github.com/AcalaNetwork/chopsticks/blob/7703263adb6ac32d675c30ee3e3ea7bbe8f26cb0/executor/src/task.rs#L144

should propagate the error instead of unwrap

Screenshot 2024-05-02 at 9 27 55 PM

this will crash the program, but it should just return an error response.

qiweiii:
fix #748

### DIFF
--- a/executor/src/task.rs
+++ b/executor/src/task.rs
@@ -133,7 +133,9 @@ pub async fn run_task(task: TaskCall, js: crate::JsCallback) -> Result<TaskRespo
     let mut runtime_logs: Vec<LogInfo> = vec![];
 
     for (call, params) in task.calls {
-        let vm_res = runtime_call::run(runtime_call::Config {
+        log::trace!("Calling {}", call);
+
+        let vm = runtime_call::run(runtime_call::Config {
             virtual_machine: vm_proto.clone(),
             function_to_call: call.as_str(),
             parameter: params.into_iter().map(|x| x.0),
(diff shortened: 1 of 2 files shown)


executor/src/task.rs
```diff
@@ -133,18 +133,23 @@ pub async fn run_task(task: TaskCall, js: crate::JsCallback) -> Result<TaskRespo
...
             function_to_call: call.as_str(),
             parameter: params.into_iter().map(|x| x.0),
             storage_main_trie_changes,
             max_log_level: task.runtime_log_level,
             calculate_trie_changes: false,
-        })
-        .unwrap();
+        });
 
         log::trace!("Calling {}", call);
```
xlc, line 145: move the log up before run. there isn't any reason it must be called after run and better to keep the result handling code together

executor/src/task.rs
```diff
@@ -133,18 +133,23 @@ pub async fn run_task(task: TaskCall, js: crate::JsCallback) -> Result<TaskRespo
...
 
+        if let Err((start_err, _)) = vm_res {
+            return Ok(TaskResponse::Error(start_err.to_string()));
+        }
+
+        let mut vm = vm_res.unwrap();
```
xlc, line 151: should always avoid unwrap if possible

in this case, it can be written as
```rs
let vm = match vm {
  Ok(vm) => vm,
  Err((start_err, _)) => return Ok(TaskResponse::Error(start_err.to_string()));
}
```

xlc: and what is the `_` here? why we should ignore it? better to write it as `_xxx` so we know what it is and add a common if needed to explain why it can be ignored

xlc (approved):

PR #752, opened 2024-05-14
`runtime-log-level: 5` for most chains

spazcoin:
This was very valuable when debugging XCM error messages. It was nice that it was the default already for Centrifuge, so seems like a good idea to make it the default for several other chains as well.


xlc (approved):

xlc:
yeah I agree

Issue #754, opened 2024-05-15
`Host function not implemented` when forking v1.11 chain. 


metricaez:
We are currently facing an issue when forking a chain that is at v1.11 with the following log: 

```
[12:09:46.874] INFO (block-builder): Failed to apply extrinsic {} Error: Host function not implemented: ext_storage_proof_size_storage_proof_size_version_1
    app: "chopsticks"
```

We have encountered a similar issue in the past in our own implementation when upgrading from `v1.8` to `v1.9` as a new node function is implemented from that version and forward as shown [here.](https://github.com/paritytech/polkadot-sdk/blob/3c3d6fceb82372a3019b37117aa453d564b212de/templates/parachain/node/src/service.rs#L43C1-L46C4)

xlc:
it needs to be implemented on smoldot first

xlc:
can you provide a wasm and config file so we can later check if the implementation works?

metricaez:
hey! unfortunately I cannot provide the current rpc we are trying with but I could build a simple lean (with no business logic at runtime level) implementation so we can test this case that comes from the node part and share it with you if that makes sense.

xlc:
you can upload a wasm and config file here that can repreoduce this issue
also give https://github.com/AcalaNetwork/chopsticks/pull/755 a try to see if it fixes the issue

metricaez:
Hi, I have built and tested #755 and it does fix the issue, thanks!

On the other hand, I am curious and a little bit confused about the wasm request since the issue comes from the node impl and not the runtime itself. Could you please elaborate ?

xlc:
issue is always coming from wasm because that’s the only thing chopsticks is testing
for this particular case, the wasm is using a new host function that wasn’t implemented by smoldot so I requested smoldot to implement it and have chopsticks updated to include the new host function

xlc:
should be fixed by #755

PR #755, opened 2024-05-17
update smoldot

qiweiii:
https://github.com/AcalaNetwork/chopsticks/issues/754

update smoldot to include ext_storage_proof_size_storage_proof_size_version_1

### DIFF
--- a/executor/src/task.rs
+++ b/executor/src/task.rs
@@ -67,6 +67,7 @@ pub struct TaskCall {
     mock_signature_host: bool,
     allow_unresolved_imports: bool,
     runtime_log_level: u32,
+    storage_proof_size: u64,
 }
 
 #[derive(Serialize, Deserialize, Debug)]
@@ -143,7 +144,7 @@ pub async fn run_task(task: TaskCall, js: crate::JsCallback) -> Result<TaskRespo
             max_log_level: task.runtime_log_level,
             calculate_trie_changes: false,
             storage_proof_size_behavior:
-                runtime_call::StorageProofSizeBehavior::ConstantReturnValue(0),
+                runtime_call::StorageProofSizeBehavior::ConstantReturnValue(task.storage_proof_size),
         });
 
         let mut vm = match vm {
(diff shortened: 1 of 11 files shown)


executor/src/task.rs
```diff
@@ -142,6 +142,8 @@ pub async fn run_task(task: TaskCall, js: crate::JsCallback) -> Result<TaskRespo
             storage_main_trie_changes,
             max_log_level: task.runtime_log_level,
             calculate_trie_changes: false,
+            storage_proof_size_behavior:
+                runtime_call::StorageProofSizeBehavior::ConstantReturnValue(0),
```
xlc, line 146: better to make this configurable from js side

qiweiii:
```rs
storage_proof_size_behavior:
         runtime_call::StorageProofSizeBehavior::ConstantReturnValue(task.storage_proof_size),
```
sth like this?

packages/core/src/blockchain/index.ts
```diff
@@ -49,6 +49,11 @@ export interface Options {
   maxMemoryBlockCount?: number
   /** Whether to process queued messages */
   processQueuedMessages?: boolean
+  /**
+   * Storage Proof Size passed to runtime.
+   * See <https://github.com/smol-dot/smoldot/blob/26dedea00eafa7073ca1d4678277272d807b3d89/lib/src/executor/host.rs#L248>
+   */
+  storageProofSize?: number
```
xlc, line 56: no need to put it here. this should be a per call option and defaults to 0 and for now the only way to set it to some value is from using the API directly

qiweiii: which api do you refer to here? and how do user pass in this value

packages/core/src/wasm-executor/index.ts
```diff
@@ -113,16 +122,10 @@ export const createProof = async (nodes: HexString[], updates: [HexString, HexSt
...
-    wasm: HexString
-    calls: [string, HexString[]][]
-    mockSignatureHost: boolean
-    allowUnresolvedImports: boolean
-    runtimeLogLevel: number
-  },
-  callback: JsCallback = emptyTaskHandler,
-) => {
+export const runTask = async (task: TaskCall, callback: JsCallback = emptyTaskHandler) => {
+  if (!task.storageProofSize) {
```
xlc, line 126:
```js
let task2 = {
  ...task,
  storageProofSize: task.storageProofSize ?? 0
}
```

avoid modify input parameter

PR #764, opened 2024-05-29
Prefetch storage

### LINKED ISSUE #759: rpc/subcommand to download and cache values for specified key ranges
useful to test a big migration touching a lot of keys

in config file, support `prefetch-storages `

```yml
prefetch-stroages:
  - 0x123456 # fetch all storages with this prefix
  - Balances # fetch all storages under Balances pallet
  - System: Account # fetch all storages under System.Account stroage
  - Tokens:
      Accounts: [5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY] # fetch all storages for Tokens.Accounts(Alice)
  - Tokens.Accounts: [5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY, { token: DOT }] # fetch this particular storage
```

It should start the chopsticks instance and fetch the storages at the same time

a new subcommand `fetch-storages` to only fetch and cache the storages
it should accept a list of string arguments, which could be
- hex: fetch all storages with this prefix
- PalletName: fetch all storages for this pallet
- PalletName.StorageName: fetch all storages for this storage

qiweiii:
close #759


packages/chopsticks/src/plugins/fetch-storage/cli.ts
```diff
@@ -20,36 +18,19 @@ export const cli = (y: Argv) => {
...
-        } else {
-          throw new Error(`Invalid block number or hash: ${config.block}`)
-        }
-
-        await fetchStorage({
-          blockHash,
+        await fetchStorages({
+          block: config.block,
+          endpoint: config.endpoint,
           dbPath: config.db ?? 'db.sqlite',
```
ermalkaleci, line 28: `config.db` needs to be required and alert users to define the one they have on config file. We're defaulting to `db.sqlite` but some configs may not use that. I would suggest we make db name to be block number. i.e: `db-555.sqlite` so you know when pointing to this db contains snapshots of block 555

xlc: yeah we can have the default name to be `db-${network_name}-${block_number}.sqlite` and have a log message print something like `Storages are saved at db-acala-12345.sqlite, use '--db=db-acala-12345.sqlite --block=12345' to apply it`

packages/chopsticks/src/plugins/fetch-storage/cli.ts
```diff
@@ -0,0 +1,35 @@
+import { z } from 'zod'
+import _ from 'lodash'
+import type { Argv } from 'yargs'
+
+import { configSchema, getYargsOptions } from '../../schema/index.js'
+import { fetchStorages } from '../../utils/fetch-storages.js'
+
+const schema = z.object({
+  ..._.pick(configSchema.shape, ['endpoint', 'block', 'db']),
+})
```
xlc, line 10:
```suggestion
const schema = z.object(
  _.pick(configSchema.shape, ['endpoint', 'block', 'db']),
)
```

xlc (approved):

xlc (approved):

PR #774, opened 2024-06-03
ci with depcheck

### LINKED ISSUE #772: check for missing deps in CI
looks like is due to https://github.com/AcalaNetwork/chopsticks/pull/764/files#diff-5ee6262753c038ef7e0ba7f03e1fc0d8440c717644a06e32a1ab1f8c9afafdf1R2
I guess we need some extra CI checks

_Originally posted by @xlc in https://github.com/AcalaNetwork/chopsticks/issues/770#issuecomment-2144668108_

qiweiii:
- check deps with `depcheck`
- fixed some missing/unused deps

close #772

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,6 @@
 		"@typescript-eslint/parser": "^7.3.1",
 		"@vitest/coverage-v8": "^1.4.0",
 		"@vitest/ui": "^1.4.0",
-		"depcheck": "^1.4.7",
 		"eslint": "^8.57.0",
 		"eslint-config-prettier": "^9.1.0",
 		"eslint-import-resolver-typescript": "^3.6.1",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -28,7 +28,8 @@
 		"docs:dev": "yarn docs:prep && vitepress dev",
 		"docs:build": "yarn docs:prep && vitepress build",
 		"docs:preview": "yarn docs:build && vitepress preview",
-		"publint": "sh ./publint.sh"
+		"publint": "sh ./publint.sh",
+		"depcheck": "yarn workspaces foreach --all -pvit run depcheck"
```
ermalkaleci, line 32: will be better to use npx like publint to avoid installing and having this as dependency

xlc (approved):

PR #787, opened 2024-07-05
Bump deps

xlc:


ermalkaleci:
@xlc updating swc will required to update build cmd

ermalkaleci:
on it

PR #790, opened 2024-07-10
fix block hash

xlc:
previously block hash is always the same as only the header hash is considered and the extrinsicRoot was mocked to zeros


packages/core/src/blockchain/block-builder.ts
```diff
@@ -341,9 +342,12 @@ export const buildBlock = async (
     callbacks?.onPhaseApplied?.('finalize', resp)
   }
 
-  const blockData = registry.createType('Block', {
-    header,
-    extrinsics: includedExtrinsic,
+  const allExtrinsics = [...inherents, ...includedExtrinsic]
+
+  const mockExtrinsicRoot = blake2AsU8a(u8aConcat(...allExtrinsics), 256)
+  const header2 = registry.createType<Header>('Header', {
```
ermalkaleci, line 348:
```suggestion
  const finalHeader = registry.createType<Header>('Header', {
```

packages/e2e/src/block.test.ts
```diff
@@ -24,4 +24,24 @@ describe('block', async () => {
 
     await Promise.all(promises)
   })
+
+  it('block hash are unique', async () => {
+    const { alice } = testingPairs()
+    await dev.setStorage({
+      System: {
+        Account: [
+          [[alice.address], { data: { free: 10 * 1e12 } }],
```
ermalkaleci, line 33:
```suggestion
          [[alice.address], { providers: 1, data: { free: 10 * 1e12 } }],
```

ermalkaleci (approved):

PR #792, opened 2024-07-11
separate Bifrost-Polkadot config from KSM Bifrost-Polkadot

spazcoin:
Bifrost Kusama and Bifrost Polkadot chains need different configs. Previously, running with `bifrost.yml` would emulate Kusama Bifrost parachain, but had token config for Polkadot tokens. Therefore split them and clean that up.

Tested by running Chopsticks and using p.js to verify tokens in the account.

### DIFF
--- a/configs/bifrost-polkadot.yml
+++ b/configs/bifrost-polkadot.yml
@@ -1,4 +1,4 @@
-endpoint: wss://bifrost-parachain.api.onfinality.io/public-ws
+endpoint: wss://bifrost-polkadot.api.onfinality.io/public-ws
 mock-signature-host: true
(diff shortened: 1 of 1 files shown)


configs/bifrost-polkadot.yml
```diff
@@ -0,0 +1,108 @@
+endpoint: wss://bifrost-polkadot.api.onfinality.io/public-ws
+mock-signature-host: true
+block: ${env.BIFROST_BLOCK_NUMBER}
```
xlc, line 3:
```suggestion
block: ${env.BIFROSTPOLKADOT_BLOCK_NUMBER}
```

xlc (approved):

PR #793, opened 2024-07-18
Add Enjin networks to configs

leonardocustodio:


leonardocustodio:
Thank you for the quick merge 😄

Issue #801, opened 2024-08-14
Chopsticks client is not compatible with Papi


Ad96el:
## Description

When attempting to create a client instance in PAPI, Chopsticks throws the following error:

```
ERROR (ws): Invalid request: {"jsonrpc":"2.0","id":"1-1","method":"chainHead_v1_follow","params":[true]}
app: "chopsticks"
```

After some investigation, it appears that the desired RPC call is not available after forking.

## Steps to Reproduce

1. Connect to any network using Polkadot.js and verify if the `chainHead_v1_follow` method is available under **Developer > RPC calls > rpc > methods**.
2. Fork the network.
3. Check if the method is still listed. It should no longer be available.

ermalkaleci:
Chopsticks is a client itself and it and has its own methods, this one is not implemented

xlc:
yeah we need to implement those new methods

Ad96el:
@xlc, thanks for the response. Could you estimate which version will include the fix?

xlc:
Unfortunately, I can't provide an estimate as we don't have extra resources atm
However happy to mentor and review if anyone is interested working on this

rflechtner:
There are ways for papi to work with clients that do not yet implement these rpc methods (https://papi.how/requirements#polkadot-sdk-110--x--1110) - as long as the client at least acts as a client on polkadot api > v1.1.0 would. We're still running into issues with this method though, which strangely results in the following error message from the chopsticks client:

```
Invalid request: {"jsonrpc":"2.0","id":"__INTERNAL_ID","method":"rpc_methods","params":[]}
``` 

Any idea what the issue may be here? May it be that the id coming from papi is throwing it off?

ermalkaleci:
id needs to be number

voliva:
Mhh I see we have to document this better, but Polkadot-API is built on top of the new [Polkadot JSON-RPC spec](https://paritytech.github.io/json-rpc-interface-spec/api/chainHead.html), and requires `chainHead`, `rpc` and `transaction` group of functions.

The compatibility layer from `polkadot-api/polkadot-sdk-compat` fixes some implementation bugs of this spec that were present in older versions of Polkadot-SDK, but that we can easily fix on the client-side. But the RPC methods defined in the new specification must be there.

Another issue might be this:

> id needs to be number

[...]

PR #807, opened 2024-08-23
make json-rpc server id handling spec-compliant

voliva:
Initial bit of work for #801 

This makes id handling spec-compliant by [JSON-RPC 2.0](https://www.jsonrpc.org/specification). Notable changes:

- The `id` can now be parsed as a number, string, null or undefined
- If the `id` is a number, it must be an integer.
- The `id` from error responses is now correct:
  - If it was while parsing the message, then it's `null`.
  - If it was while handling a request, then it's the same request id as that request.
- It handles notifications correctly: if the `id` is undefined, no response is returned. 

~~I couldn't find a good place to add tests for these cases - please let me know if you want me to add some, or if you need anything else.~~
I could find the test file, `e2e/src/http.test.ts`

### DIFF
--- a/packages/chopsticks/src/server.ts
+++ b/packages/chopsticks/src/server.ts
@@ -93,14 +93,13 @@ export const createServer = async (handler: Handler, port: number) => {
 
   const safeHandleRequest = async (request: z.infer<typeof singleRequest>) => {
     try {
-      const result = handler(request, emptySubscriptionManager)
+      const result = await handler(request, emptySubscriptionManager)
       return request.id === undefined ? undefined : { id: request.id, jsonrpc: '2.0', result }
     } catch (err: any) {
       return {
         jsonrpc: '2.0',
         id: request.id,
         error: {
-          code: -32603,
           message: err.message,
         },
       }
@@ -144,7 +143,6 @@ export const createServer = async (handler: Handler, port: number) => {
           jsonrpc: '2.0',
           id: null,
           error: {
-            code: -32700,
             message: err.message,
           },
         }),
--- a/packages/chopsticks/src/server.ts
+++ b/packages/chopsticks/src/server.ts
@@ -9,7 +9,7 @@ const httpLogger = defaultLogger.child({ name: 'http' })
 const wsLogger = defaultLogger.child({ name: 'ws' })
 
 const singleRequest = z.object({
-  id: z.number(),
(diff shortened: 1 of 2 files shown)


packages/chopsticks/src/server.ts
```diff
@@ -9,7 +9,7 @@ const httpLogger = defaultLogger.child({ name: 'http' })
 const wsLogger = defaultLogger.child({ name: 'ws' })
 
 const singleRequest = z.object({
-  id: z.number(),
+  id: z.optional(z.union([z.number().int(), z.string(), z.null()])),
```
ermalkaleci, line 12: why does this need to be null and undefined? don't we need either number or string?

voliva: For polkadot-api we would only need it to accept strings, but I chose to make it spec-compliant instead. The spec for the id field on a request says:

> An identifier established by the Client that MUST contain a String, Number, or NULL value if included. If it is not included it is assumed to be a notification. The value SHOULD normally not be Null and Numbers SHOULD NOT contain fractional parts
> [...]
> A Notification is a Request object without an "id" member. A Request object that is a Notification signifies the Client's lack of interest in the corresponding Response object, and as such no Response object needs to be returned to the client. The Server MUST NOT reply to a Notification, including those that are within a batch request.

Meaning undefined (not included) and null have different meaning: Not included means the request is a notification, which means it's not expecting any response back, whereas having a `null` id requires the server to reply with to the message (with the same null value for the id)

If you rather have a smaller change even if it's not spec-compliant I'm ok simplifying it back, but I thought this would have more value.

packages/chopsticks/src/server.ts
```diff
@@ -112,24 +127,21 @@ export const createServer = async (handler: Handler, port: number) => {
 
       let response: any
       if (Array.isArray(parsed.data)) {
-        response = await Promise.all(
-          parsed.data.map((req) => {
-            const result = handler(req, emptySubscriptionManager)
-            return { id: req.id, jsonrpc: '2.0', result }
-          }),
-        )
+        response = await Promise.all(parsed.data.map(safeHandleRequest))
```
ermalkaleci, line 130:
```suggestion
        response = await Promise.all(parsed.data.map(safeHandleRequest).filter((r) => r !== undefined))
```

ermalkaleci (approved):
LGTM

PR #810, opened 2024-08-25
Add one more condition to check whether the parachain is async backing chain

sfffaaa:
Support the slot duration if the aura didn't expose

Based on the Parity team's async implementation doc, the meta.consts.timestamp.minimumPeriod == 0 should represent async backing.

https://wiki.polkadot.network/docs/maintain-guides-async-backing


ermalkaleci:
I'm fixing this on https://github.com/AcalaNetwork/chopsticks/pull/809
Slot duration should come from aura

PR #813, opened 2024-08-26
JSON RPC interface spec: chainHead_v1

voliva:
More work for #801

This is the most important group of functions to be implemented from the spec. This makes PAPI already compatible except to send transactions (which needs the `transaction_v1` group of functions).

It's not fully spec-compliant, specifically for the `chainHead_v1_storage` call, which has 2 variations that are more complex (`closestDescendantMerkleValue`, `hash`,  `descendantsHashes`, and dealing with `childTrie`), but none of those features are used by PAPI. I rather make small incremental PRs, so I would initially focus on having full PAPI compatibility, and later on we can make it full spec compliant.

I'd also like to add tests, what do you prefer? Should I raise them as a separate PR or would you like it in this one? I'm thinking on keep raising incremental PRs, first I'll work on the tests for `chainHead_v1`, then implement `transaction_v1` and its tests, and lastly `chainSpec_v1` (the last group used by PAPI, but the least important).


ermalkaleci:
Please add tests into this PR

ermalkaleci:
> @ermalkaleci I see that the e2e tests use PolkadotJS to test everything works as expected, but PolkadotJS doesn't use the new RPC-interface
> 
> Is it alright if I add polkadot-api (or a lower level sub-package) as a devDependency to run the e2e tests for the new rpc spec?

of course, go ahead

xlc:
you could add a new env for papi tests if it helps

packages/e2e/src/helper.ts
```diff
@@ -102,13 +104,17 @@ export const setupAll = async ({
         noInitWarn: true,
       })
 
+      const substrateClient = createClient(getWsProvider(`ws://localhost:${port}`))
```
xlc, line 107: I am not sure how much overhead is there to create a papi client because this is used by all tests and papi client will only be used by some tests. Maybe add an option to create papi client when needed?

packages/e2e/src/helper.ts
```diff
@@ -102,13 +104,17 @@ export const setupAll = async ({
...
+
       await apiPromise.isReady
 
       return {
         chain,
         ws,
         api: apiPromise,
+        substrateClient,
         async teardown() {
+          substrateClient.destroy()
```
xlc, line 117: just to confirm, no need to await here?

voliva: Yep, it's synchronous.

packages/e2e/src/helper.ts
voliva: There shouldn't be too much overhead other than a new WS connection opened - `@polkadot-api/substrate-client` doesn't do any operations by its own.

I can move it to a separate function or add the option flag if it's still a concern 👍

voliva: I've been adding the transaction_v1 and when writing the tests I found out it would be great to actually have the v15 metadata, because then we can use a higher-level library to track transactions and check for success, invalid and failed.

I'll take this chance to put it as a separate env that only gets instantiated in the new RPC spec tests (on this PR)

xlc: can you put this back to draft and make it ready when all the work is done so I know when it is a good time to review again?

voliva: Sorry I missed this - It's now ready 👍

xlc (approved):

PR #818, opened 2024-09-04
no auto block in manual mode

xlc:
when block building mode is manual, it should be manual

### DIFF
--- a/packages/core/src/blockchain/txpool.ts
+++ b/packages/core/src/blockchain/txpool.ts
@@ -197,24 +197,27 @@ export class TxPool {
         unsafeBlockHeight,
       })
 
-      // with the latest message queue, messages are processed in the upcoming block
-      if (!this.#chain.processQueuedMessages) return
-      // if block was built without horizontal or downward messages then skip
-      if (_.isEmpty(horizontalMessages) && _.isEmpty(downwardMessages)) return
+      // try to process queued messages if not in manual mode
+      if (this.#mode !== BuildBlockMode.Manual) {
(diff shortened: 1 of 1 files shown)


xlc:
@ermalkaleci Moonbeam and Astar is currently always having messageQueue.bookStateFor(Parent) so unconditionally builds extra block when received messages and that breaks e2e tests. Not sure why that is happening but this will be an easy fix to get the e2e tests working with some tweaks

ermalkaleci:
@xlc I disabled it auto processing https://github.com/AstarNetwork/e2e-tests/pull/70/files#diff-aa74785dd9e1c615d463ffed54f55bb20f72f4e5dadccc2d7b61a250b0907b29R19-R21
I will take a look today

ermalkaleci:
We have this for Parent, I don't know if it's a bug or not. Chopsticks will check the size which is gt 0 and build although we handle queued messages on idle
```
[
      Parent
]
{
      begin: 57
      end: 57
      count: 1
      readyNeighbours: null
      messageCount: 1
      size_: 11
 }
 ```

ermalkaleci:
@xlc I think rather than disabled it on manual mode ( I think it should cause other problems), we instead check `end > begin`. I will solve current issue

xlc:
> @xlc I think rather than disabled it on manual mode ( I think it should cause other problems), we instead check `end > begin`. I will solve current issue

that will be better. you want to make a PR?

ermalkaleci:
@xlc you can just update this one

ermalkaleci (approved):

PR #819, opened 2024-09-09
feat: add transaction_v1 group of functions

voliva:
This PR adds the `transaction_v1` group of functions of the JSON RPC interface spec into the RPC server.

It's fairly simple, since it's only two functions: one starting a broadcast, and the other to stop it.

We'll be missing only one last group to make it fully compatible with papi, the `chainSpec` one. Looks fairly simple, will work on it next and raise as a separate PR.

### DIFF
--- a/packages/core/src/rpc/rpc-spec/transaction_v1.ts
+++ b/packages/core/src/rpc/rpc-spec/transaction_v1.ts
@@ -1,0 +1,33 @@
+import { Handler } from '../shared.js'
+import { HexString } from '@polkadot/util/types'
+import { defaultLogger } from '@acala-network/chopsticks-core'
+
+const logger = defaultLogger.child({ name: 'rpc-transaction_v1' })
+const randomId = () => Math.random().toString(36).substring(2)
+
+/**
+ * Submit the extrinsic to the transaction pool
+ *
+ * @param context
+ * @param params - [`extrinsic`]
+ *
+ * @return operation id
+ */
+export const transaction_v1_broadcast: Handler<[HexString], string | null> = async (context, [extrinsic]) => {
+  await context.chain.submitExtrinsic(extrinsic).catch((err) => {
+    // As per the spec, the invalid transaction errors should be ignored.
+    logger.warn('Submit extrinsic failed', err)
+  })
+
+  return randomId()
+}
+
+/**
+ * Stop broadcasting the transaction to other nodes.
+ *
+ */
+export const transaction_v1_stop: Handler<[string], null> = async (_context, [_operationId]) => {
+  // Chopsticks doesn't have any process to broadcast the transaction through P2P
+  // so stopping doesn't have any effect.
+  return null
+}
--- a/packages/core/src/rpc/rpc-spec/transaction_v1.ts
+++ b/packages/core/src/rpc/rpc-spec/transaction_v1.ts
@@ -1,6 +1,6 @@
 import { Handler } from '../shared.js'
 import { HexString } from '@polkadot/util/types'
-import { defaultLogger } from '@acala-network/chopsticks-core'
+import { defaultLogger } from '../../logger.js'
 
 const logger = defaultLogger.child({ name: 'rpc-transaction_v1' })
 const randomId = () => Math.random().toString(36).substring(2)
--- a/packages/e2e/src/rpc-spec.test.ts
+++ b/packages/e2e/src/rpc-spec.test.ts
@@ -1,6 +1,8 @@
 import { ApiPromise } from '@polkadot/api'
+import { RuntimeContext } from '@polkadot-api/observable-client'
 import { describe, expect, it } from 'vitest'
 import { dev, env, observe, setupPolkadotApi, testingPairs } from './helper.js'
+import { firstValueFrom } from 'rxjs'
 
 const testApi = await setupPolkadotApi(env.acalaV15)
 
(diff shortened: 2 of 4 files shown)


packages/e2e/src/rpc-spec.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    const tx = await api.tx.balances.transferKeepAlive(bob.address, 100n).signAsync(alice)
+    const { nextValue, subscription } = observe(chainHead.trackTx$(tx.toHex()))
+    const resultPromise = nextValue()
+    const broadcast = testApi.observableClient.broadcastTx$(tx.toHex()).subscribe()
+
+    // We don't have a confirmation of when the transaction has been broadcasted through the network
+    // it just continues to get broadcasted through the nodes until we unsubscribe from it.
+    // In this case, where there's only one node, a couple of blocks should be enough.
+    await dev.newBlock()
+    const hash = await dev.newBlock()
```
ermalkaleci, line 24: tx is included immediately, why do we need 2 blocks?

packages/core/src/rpc/rpc-spec/transaction_v1.ts
```diff
@@ -0,0 +1,33 @@
+import { Handler } from '../shared.js'
+import { HexString } from '@polkadot/util/types'
+import { defaultLogger } from '@acala-network/chopsticks-core'
```
ermalkaleci, line 3: import from logger.js instead

packages/core/src/rpc/rpc-spec/transaction_v1.ts
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @return operation id
+ */
+export const transaction_v1_broadcast: Handler<[HexString], string | null> = async (context, [extrinsic]) => {
+  await context.chain.submitExtrinsic(extrinsic).catch((err) => {
+    // As per the spec, the invalid transaction errors should be ignored.
+    logger.warn('Submit extrinsic failed', err)
+  })
+
+  return randomId()
```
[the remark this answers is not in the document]
voliva: Yes, according to the [spec](https://paritytech.github.io/json-rpc-interface-spec/api/transaction_v1_broadcast.html):

> Return value: String representing the operation, or null if the maximum number of broadcasted transactions has been reached.

The maximum number of broadcasted transactions must be at least 4, but doesn't specify any upper limit. In this case as it doesn't take resources, I think we can do unlimited.

packages/e2e/src/rpc-spec.test.ts
voliva: polkadot-api synchronously sends the broadcast, but that goes through the websocket connection, which is async. Unfortunately we don't have a confirmation when the broadcast is completed (since the node just keeps broadcasting it to every node).

Realistically, I could wait a fixed amount of time (I saw ~250ms is enough to have the transaction included in the tx pool), but waiting for two blocks felt a bit closer to what happens in a real chain too.

packages/e2e/src/rpc-spec.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    broadcast.unsubscribe()
+    subscription.unsubscribe()
+    chainHead.unfollow()
+  })
+})
+
+const UPGRADED = 0x80000000_00000000_00000000_00000000n
+const INITIAL_ACCOUNT_VALUE = 100_000_000_000_000n
+async function prepareChainForTx() {
+  const api = await ApiPromise.create({
```
ermalkaleci, line 42: One more thing and it's ready
You don't need all this. `await setup()` will make it's ready.
Also you can use `await setStorage(chain, storage)` without the need of dev, just in case you need it

packages/e2e/src/rpc-spec.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    noInitWarn: true,
+  })
+  await api.isReady
+  await dev.setStorage({
+    System: {
+      Account: [
+        [
+          [alice.address],
+          {
+            data: { free: INITIAL_ACCOUNT_VALUE, flags: UPGRADED },
```
ermalkaleci, line 53: isn't providers required?

packages/e2e/src/rpc-spec.test.ts
voliva: From the previous PR #813 I split up this function into two different `setup`s, one for PolkadotJS, and another for polkadot-api, since xlc was worried that having two clients [would have too much overhead](https://github.com/AcalaNetwork/chopsticks/pull/813#discussion_r1742786765) to have them on all tests.

So only for transactions I thought having this utility to set up the chain for these specific tests.

packages/e2e/src/rpc-spec.test.ts
voliva: Oh right! with providers I don't need the `flags` either, since the upgrade happens automatically. Thank you!

packages/e2e/src/rpc-spec.test.ts
```diff
@@ -0,0 +1,82 @@
...
+const { alice, bob } = testingPairs()
+
+describe('transaction_v1', async () => {
+  it('sends and executes transactions', async () => {
+    const chainHead = testApi.observableClient.chainHead$()
+
+    const api = await prepareChainForTx()
+
+    const TRANSFERRED_VALUE = 100n
+    const tx = await api.tx.balances.transferKeepAlive(bob.address, TRANSFERRED_VALUE).signAsync(alice)
```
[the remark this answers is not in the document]
voliva: papi currently can only build transactions on the top-level client, which needs a codegen step to generate descriptors. This is because we can generate the call data with the metadata-builders package, but the signed extensions and signature is only handled in the top-level client. @josepot is working on a [new sub-package](https://github.com/polkadot-api/polkadot-api/pull/673) that will make it possible to create them without the top-level client, but it will take a while.

So I thought that I could create the transactions and sign them with pjs without actually sending them, and then broadcast and track them with polkadot-api (which is the part that uses the new rpc interface spec)

packages/e2e/src/rpc-spec.test.ts
ermalkaleci: I can see now, didn't check tx signing carefully, I thought you were using papi to create the tx.

ermalkaleci (approved):

PR #820, opened 2024-09-12
feat: add chainSpec_v1 group of functions

### LINKED ISSUE #801: Chopsticks client is not compatible with Papi

voliva:
Last missing group of functions to provide support with papi, so this closes #801 

I may need help with the `genesisHash` function - I thought this would be available on the chain object directly, but it's not.

I called `getBlockHash(0)`, which seems to get the job done, it seems to work when starting from genesis and from a wsUrl, but the types say it might return null. I couldn't find any case where this happens, but I'm also not sure what to do in this case.... any ideas?

### DIFF
--- a/packages/core/src/rpc/rpc-spec/chainSpec_v1.ts
+++ b/packages/core/src/rpc/rpc-spec/chainSpec_v1.ts
@@ -1,0 +1,14 @@
+import { ChainProperties } from '../../index.js'
+import { Handler } from '../shared.js'
+
+export const chainSpec_v1_chainName: Handler<[], string> = async (context) => {
+  return context.chain.api.getSystemChain()
+}
+
+export const chainSpec_v1_genesisHash: Handler<[], string> = async (context) => {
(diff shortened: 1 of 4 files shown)


xlc:
You can call getBlockHash with a future block number and it will return null, however, for genesis, it should always give you a block hash. It will be an unreachable code path for the null case. You could throw error like "Unexpected null genesis" if it ever happens.

xlc (approved):

Issue #757, opened 2024-05-20
Override relay storage keys included in set_validation_data inherent


RomarQ:
The storage `paras > upgradeCooldowns` is part of the relay storage keys included with `set_validation_data` inherent. 

Is there currently a way to remove relay storages keys when testing a parachain with chopsticks?

ermalkaleci:
https://github.com/paritytech/polkadot-sdk/blob/313fe0f9a277f27a4228634f0fb15a1c3fa21271/cumulus/client/parachain-inherent/src/lib.rs#L99-L123 I don't think it is included

RomarQ:
You are right, it is the `upgrade_restriction_signal` storage.

Is there a way in chopsticks to change the value for that storage key?

xlc:
it should be possible to add a new option to `dev_newBlock` to pass some parameters to the inherent builder

ermalkaleci:
https://github.com/AcalaNetwork/chopsticks/blob/15ef2280951b1c93baa9239f4b32721b2085afff/packages/core/src/blockchain/inherent/parachain/validation-data.ts#L260-L269
this is how we manipulate relay-chain data

RomarQ:
But can I override those externally without changing the in the chopsticks internals?

RomarQ:
Ok, thanks.

PR #824, opened 2024-09-24
feat(dev_newBlock): Adds option to override relaychain state

RomarQ:
Adds a solution for this issue: https://github.com/AcalaNetwork/chopsticks/issues/757

The changes in this PR add a new parameter field `relayChainStateOverrides` to `dev_newBlock` RPC. The objective of this new field is to give more flexibility to the user and allow custom relay chain state when creating new blocks.

### DIFF
--- a/packages/core/src/blockchain/inherent/parachain/validation-data.ts
+++ b/packages/core/src/blockchain/inherent/parachain/validation-data.ts
@@ -98,7 +98,7 @@ export class SetValidationData implements InherentProvider {
 
     const extrinsic = await getValidationData(parent)
 
-    const newEntries: [HexString, HexString | null][] = []
+    let newEntries: [HexString, HexString | null][] = []
     const downwardMessages: DownwardMessage[] = []
     const horizontalMessages: Record<number, HorizontalMessage[]> = {}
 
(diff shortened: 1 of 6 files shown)


packages/core/src/blockchain/inherent/parachain/validation-data.ts
```diff
@@ -268,6 +268,12 @@ export class SetValidationData implements InherentProvider {
       newEntries.push([upgradeKey, null])
     }
 
+    // Apply relay state overrides
+    newEntries = (params.relayChainStateOverrides || []).reduce(
```
xlc, line 272: this code is not easy to understand. wouldn't it be simpler with for loop?

packages/e2e/src/build-parachain-block.test.ts
```diff
@@ -0,0 +1,50 @@
+import { afterAll, describe, expect, it } from 'vitest'
+
+import { TypeRegistry } from '@polkadot/types'
+import { decodeProof } from '@acala-network/chopsticks-core'
+import { setupAll } from './helper.js'
+import { upgradeRestrictionSignal } from '@acala-network/chopsticks-core/utils/proof.js'
+
+describe.runIf(process.env.CI || process.env.RUN_ALL).each([
```
ermalkaleci, line 8: there's no need for this many tests
A single chain will do
```suggestion
describe('override-relay-state-proof', async () => {
  const { chain, dev, teardown, api } = await networks.acala()
  ...
```

packages/core/src/blockchain/inherent/parachain/validation-data.ts
RomarQ: Pushed a more imperative version of the code. 

I usually use reduce where possible to make the code more compact and simpler sometimes.

packages/e2e/src/build-parachain-block.test.ts
RomarQ: Applied your suggestions

xlc (approved):

PR #838, opened 2024-10-15
feat(chainHead_v1): operationWaitingForContinue and chainHead_v1_continue

### LINKED ISSUE #837: PAPI storage getEntries() returns only 100 entries
Hi,

First of all thank you for providing us with this great tool! 

While trying to run my polkadot-api (PAPI) based dapp on chopsticks I noticed that on Asset Hub, `api.query.Assets.Metadata.getEntries()` returns only the first 100 items, instead of 300+

I noticed that it can be fixed by wrapping the provider with `withPolkadotSdkCompat`, but I've been told that it shouldn't be necessary.

Repo to reproduce : https://github.com/0xKheops/papi-chopsticks-getentries

I get this output:
image

voliva:
I've added pagination by performing storage queries with 1000 values at a time, then storing the pending requests on the "callbacks" map we had, which I have repurposed to better represent the state of each follow subscription.

I also added a test by querying a storage query that has more than 1000 values, but without being too much. I've found System.BlockHash has 1201 entries, so I thought that's a good candidate. Let me know if I should change this for a better alternative.

Closes #837

### DIFF
--- a/packages/core/src/rpc/rpc-spec/chainHead_v1.ts
+++ b/packages/core/src/rpc/rpc-spec/chainHead_v1.ts
@@ -187,7 +187,6 @@ async function getDescendantValues(
   }>
   next: DescendantValuesParams | null
 }> {
-  const pageSize = 100
   const keys = await block.getKeysPaged({
     ...params,
     pageSize: PAGE_SIZE,
@@ -202,7 +201,7 @@ async function getDescendantValues(
     ),
   )
 
-  if (keys.length < pageSize) {
+  if (keys.length < PAGE_SIZE) {
     return {
       items,
       next: null,
@@ -213,7 +212,7 @@ async function getDescendantValues(
     items,
     next: {
       ...params,
-      startKey: keys[pageSize - 1],
+      startKey: keys[PAGE_SIZE - 1],
     },
   }
 }
--- a/packages/core/src/rpc/rpc-spec/chainHead_v1.ts
+++ b/packages/core/src/rpc/rpc-spec/chainHead_v1.ts
@@ -5,7 +5,17 @@ import { defaultLogger } from '../../logger.js'
 
 const logger = defaultLogger.child({ name: 'rpc-chainHead_v1' })
 
-const callbacks = new Map<string, (data: any) => void>()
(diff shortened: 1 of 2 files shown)


packages/core/src/rpc/rpc-spec/chainHead_v1.ts
```diff
@@ -5,7 +5,17 @@ import { defaultLogger } from '../../logger.js'
 
 const logger = defaultLogger.child({ name: 'rpc-chainHead_v1' })
 
-const callbacks = new Map<string, (data: any) => void>()
+type DescendantValuesParams = {
+  prefix: string
+  startKey: string
+}
+const following = new Map<
```
xlc, line 12: One downside about global variables is that for some use cases, we need to handle chopsticks instances running at a same time (e.g. unit tests) so need to ensure they don't interfere with each other. 
In this particular case, it is fine as everything is indexed by a unique random ID. The only drawback I can see is potential memory leak (i.e. destroy a chopsticks instance will not clear entry here). But that's not a very big deal here.
So this is ok for now. Maybe later refactor this into context somehow.

packages/core/src/rpc/rpc-spec/chainHead_v1.ts
```diff
@@ -166,6 +176,48 @@ export interface StorageItemRequest {
...
+  block: Block,
+  params: DescendantValuesParams,
+): Promise<{
+  items: Array<{
+    key: string
+    value?: HexString
+  }>
+  next: DescendantValuesParams | null
+}> {
+  const pageSize = 100
```
xlc, line 190: you have 2 page_size and one is 1000 and one is 100?

packages/core/src/rpc/rpc-spec/chainHead_v1.ts
xlc: #839

packages/core/src/rpc/rpc-spec/chainHead_v1.ts
voliva: my bad - leftover from previous code.

Thanks for pointing it out!

PR #843, opened 2024-10-24
fix node server localhost

### LINKED ISSUE #842: Connection failling version 0.16.2 MacOS
Running chopsticks version `0.16.2` is causing a problem. I'm getting an error `Abnormal closure` i've tried to run it with the flag `LOG_LEVEL=debug` but no errors appear there. From the test that i've been conducted this also happen on version `0.16.1`. I can use version `0.9.13` to run it without any issue. This only happened to me on MacOS 15 it seems to be fine on ubuntu 24.04.

image

qiweiii:
fix #842

### DIFF
--- a/executor/package.json
+++ b/executor/package.json
@@ -1,7 +1,7 @@
 {
 	"name": "@acala-network/chopsticks-executor",
 	"description": "Chopsticks executor",
-	"version": "0.16.2",
+	"version": "0.16.3",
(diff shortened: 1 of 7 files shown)


xlc (approved):

ermalkaleci:
Will be better to add a section on README explaining this rather than forcing 127

PR #849, opened 2024-10-30
fix macos resolving localhost to ipv6 only

ermalkaleci:


xlc (approved):

xlc:
this is a CLI breaking change. do we want to make addr as alias or print a message saying please use host?
also I think we could make a 1.0 release after this.

ermalkaleci:
Yes we can, will add alias

PR #864, opened 2024-12-10
Support archive_unstable_body

### LINKED ISSUE #860: Support `archive_unstable_body`
Hello there, I'd like to add support for `archive_unstable_body` but I can't manage to find if this is an alias for a known call already. I suppose it is though. Can you let me know how to tell? I'd make a PR.

Cheers

Tbaut:
closes #860 
This adds support for `archive_unstable_body` which answers an array of extrinsics.

Please let me know what's the best way to test this without hardcoding the extrinsics


packages/e2e/src/chain.test.ts
```diff
@@ -42,6 +42,13 @@ describe('chain rpc', () => {
     await check(api.rpc.chain.getBlock(hash0)).toMatchSnapshot()
     await check(api.rpc.chain.getBlock(hash1000)).toMatchSnapshot()
 
+    expect(await api.rpc('archive_unstable_body', hash1000)).toEqual(
+      expect.arrayContaining([
+        '0x280401000be3da72cc7d01',
```
Tbaut, line 47: it's most probably not ok to hardcode these. How can I access the lower parts of a snapshot?

packages/core/src/rpc/substrate/chain.ts
```diff
@@ -71,6 +71,20 @@ export const chain_getBlock: Handler<
   }
 }
 
+/**
+ * @param context
+ * @param params - [`blockhash`]
+ *
+ * @return Block extrinsics
+ */
+export const archive_unstable_body: Handler<[HexString], HexString[]> = async (context, [hash]) => {
```
Tbaut, line 80: Would you like me to create a new `archive` folder, and put this function there?

ermalkaleci: Up to you

packages/e2e/src/chain.test.ts
ermalkaleci: It's fine. Block 100 isn't going to change

packages/core/src/rpc/substrate/chain.ts
xlc: it will make finding things a bit easier

packages/e2e/src/chain.test.ts
xlc: any reason to not use match snapshot of the whole body?

Tbaut: the thing is that `archive_unstable_body` only responds with the extrinsics array. However `toMatchSnapshot` is for the `body`, the `header` etc. So if I could have selected juste the `extrinsics` from this snaptshot, it would have been ideal.

Tbaut: thanks, I have indeed no idea, and was definitely looking for help. I'll check it out in the morning.

Tbaut: Thanks for the guidance, it's now using a snapshot

packages/core/src/rpc/substrate/chain.ts
Tbaut: done, I also moved the archive alias.

ermalkaleci (approved):

xlc (approved):

PR #866, opened 2024-12-11
Fix loopback address display for IPv6

rockbmb:
In IPv6, the loopback address is `::1`, short for `0000:0000:0000:0000:0000:0000:0000:0001`.
Before this commit, it was displayed as `::`, which is short for `0000:0000:0000:0000:0000:0000:0000:0000` - not the same addresses.

See [RFC 4219 for IPv6](https://www.rfc-editor.org/rfc/rfc4291#section-2.2).

### DIFF
--- a/packages/chopsticks/src/server.ts
+++ b/packages/chopsticks/src/server.ts
@@ -289,6 +289,11 @@ export const createServer = async (handler: Handler, port: number, host?: string
     })
   })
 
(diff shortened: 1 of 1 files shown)


xlc:
it is displaying the listen address, and in the case it is listening on the unspecified address, it should display as it is?

rockbmb:
@xlc this was a misunderstanding on my part - I assumed the behavior should be the same as when it displayed IPv4's `localhost` for use in PJS.

PR #898, opened 2025-03-27
avoid fetch storages already in db

### LINKED ISSUE #897: Prefetch storages becomes useless by background storage fetch
getKeysPaged will try to download storages and not respect prefetched storages
https://github.com/AcalaNetwork/chopsticks/blob/462fdc458f424e4cb68aeeb42825cc79299f9ca8/packages/core/src/blockchain/storage-layer.ts#L113

potential solution: check if next key is already in db and don't call getStorageBatch

xlc:
closes #897

this don't eliminate getKeysPaged call which requires we store fetch ranged in db but should significantly reduce RPC requests with a prefetched db

### DIFF
--- a/packages/core/src/blockchain/storage-layer.ts
+++ b/packages/core/src/blockchain/storage-layer.ts
@@ -35,12 +35,12 @@ export interface StorageLayerProvider {
 
 export class RemoteStorageLayer implements StorageLayerProvider {
   readonly #api: Api
-  readonly #at: string
+  readonly #at: HexString
   readonly #db: Database | undefined
   readonly #keyCache = new KeyCache(PREFIX_LENGTH)
   readonly #defaultChildKeyCache = new KeyCache(CHILD_PREFIX_LENGTH)
 
-  constructor(api: Api, at: string, db: Database | undefined) {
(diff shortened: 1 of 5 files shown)


packages/core/src/blockchain/storage-layer.ts
```diff
@@ -111,12 +111,25 @@ export class RemoteStorageLayer implements StorageLayerProvider {
...
       if (this.#db) {
-        // batch fetch storage values and save to db, they may be used later
-        this.#api.getStorageBatch(prefix as HexString, batch as HexString[], this.#at as HexString).then((storage) => {
-          for (const [key, value] of storage) {
-            this.#db?.saveStorage(this.#at as HexString, key as HexString, value)
+        // filter out keys that are not in the db]
+        const newBatch: HexString[] = []
+
+        for (const key of batch) {
+          const res = await this.#db.queryStorage(this.#at, key)
```
ermalkaleci, line 118: it will be better if we add additional api to check existing keys in batches

xlc: it is not too slow to make 1000 requests to sqlite but yeah we can improve it #899

ermalkaleci (approved):

PR #921, opened 2025-05-27
feat: add support for chainHead_v1_storage hash queries

voliva:
I think this is the last piece that was missing for `chainHead_v1` - Support for hash queries.

IIRC we postponed this because there was no direct way of getting the hashed values. What I used to solve it in this PR is to grab the value and hash it with blake2.

We recently added a feature that relies on this for papi, and with chopsticks it was always returning `null` and logging the "not implemented" warning.


xlc (approved):

ermalkaleci:
please pull master

Issue #961, opened 2025-09-07
RPC limit on getMany storage queries


voliva:
I recently opened a PR #951 to speed up fetching multiple storage values at once. @ermalkaleci warned that this could cause RPC limits, but I couldn't find any reference for those limits so we went ahead with the change.

However, recently I came up with the limit that he was probable referring to:

```json
{
  "jsonrpc":"2.0",
  "id":78,
  "error":{
    "code":-32008,
    "message":"Response is too big",
    "data":"Exceeded max limit of 15728640"
  }
}
```

So the limit it's not on the amount of queries `state_queryStorageAt` can handle (which chopsticks already limits to 1000), but rather the response can't exceed a maximum value... which means that also depends on the storage query and the actual value of that.

This happened in a storage query with 749 items for Staking.EraStakersPaged, which is a storage entry where every item is quite large.

So I'm not sure what would be the best solution to that. Well, the new JSON-RPC spec wouldn't have this problem 😅, but using the current setup what would be the best way of setting a limit without knowing how big the storage entries are?

Maybe one option is always splitting these into 2-4 requests instead? It would still be better than querying items one-by-one…

ermalkaleci:
Yes that's the limit I was referring to.
As you said the easy way is to chuck on api implementation if this error is received. Api will try with provided limit, if it fails because of size limit then it will retry with smaller limit and merge results together and return it. This way no change is required to the rest of the code

PR #962, opened 2025-09-07
fix: split getStorageBatch into smaller batches

voliva:
Addresses #961 

I've set the batch size for any remote getStorageBatch to 300, but also making each batch to get a similar size (so a 301 operation will have two batches, with sizes 151 and 150)


voliva:
> My suggestion is to handle this on api implementation, and only reduce size if it fails a first try

Good point, I thought `this.#api` was already PJS' api. My only question though is if this error is standard across RPCs. I've added a check by code or message just in case.

ermalkaleci:
@voliva can you skip that test so we can merge?

PR #1037, opened 2026-05-29
Support Bulletin chain (`pallet-transaction-storage`)

rockbmb:
## Summary

Chopsticks could not fork Bulletin on Paseo. Three things were broken.

**1. Missing inherent.** Bulletin requires a `transactionStorage.applyBlockInherents` call every block. Chopsticks did not produce one. The runtime panics in `on_finalize` if it is absent. Added an inherent provider that creates the call and forces the `ProofChecked` flag in storage — Chopsticks cannot generate real storage proofs because it does not have the underlying chunk data.

**2. `ext_trie_blake2_256_verify_proof` not implemented.** The runtime calls this to verify Merkle proofs for stored transactions. smoldot had it stubbed with a panic. Implemented all four `verify_proof` variants using smoldot existing `trie::proof_decode`.

**3. `ext_transaction_index_index` reads params wrong.** The handler assumed param 0 is an i64 pointer+size pair. It is actually three i32s. Panics on `unreachable!()` the moment a `storeWithCidConfig` extrinsic triggers transaction indexing. Same bug in `ext_transaction_index_renew`. Both are no-ops — just removed the bad param reads.

Fixes 2 and 3 are in the smoldot submodule bump, pending merge of https://github.com/paritytech/smoldot/pull/3276. CI will fail until that lands.

## Testing

Validated against `wss://paseo-bulletin-next-rpc.polkadot.io` (Paseo Bulletin Next, para 1501):

- `dev_newBlock`: 7+ consecutive blocks, zero errors
- Submitted and dispatched representative Bulletin extrinsics through the tx pool:
  - `sudo(authorizeAccount)` -> `AccountAuthorized`
  - `store` -> `Stored` + content hash + CID
  - `storeWithCidConfig` (SHA2-256, DAG-PB) -> `Stored` + CID
  - `storeWithCidConfig` (Blake2b256, raw) -> `Stored` + CID
  - `forceRenew(Position)` -> `Renewed`
  - `sudo(authorizePreimage)` -> `PreimageAuthorized`
  - `sudo(refreshAccountAuthorization)` -> `AccountAuthorizationRefreshed`
  - `sudo(refreshPreimageAuthorization)` -> `PreimageAuthorizationRefreshed`
- `run-block` replay: all extrinsics (including signed `storeWithCidConfig`) execute without crashes. `finalize_block` still fails on replay due to a state-availability limitation (the inherent proof re-verification needs historical `Transactions` data that is not accessible through RPC-backed state).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "vendor/smoldot"]
 	path = vendor/smoldot
-	url = https://github.com/smol-dot/smoldot.git
+	url = https://github.com/paritytech/smoldot.git
--- a/executor/Cargo.lock
+++ b/executor/Cargo.lock
@@ -63,6 +63,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
+name = "base32"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "022dfe9eb35f19ebbcb51e0b40a5ab759f46ad60cadf7297e0bd085afb50e076"
+
+[[package]]
 name = "base64"
 version = "0.22.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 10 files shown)


rockbmb:
Depends on https://github.com/paritytech/smoldot/pull/3276 to be merged; otherwise, the `smoldot` vendored here would need to point to a fork.

rockbmb:
https://github.com/paritytech/smoldot/pull/3276 was made redundant as https://github.com/paritytech/smoldot/pull/3263 already existed.

xlc (approved):

PR #1055, opened 2026-07-14
fix(core): serve getKeysPaged scan continuations from the KeyCache again

### LINKED ISSUE #1054: getKeysPaged bypasses the KeyCache on every scan continuation since 1.4.0 — one upstream RPC per key (~98x slower, ~358x the RPC traffic)
## Summary

Since **1.4.0**, `RemoteStorageLayer.getKeysPaged` bypasses the `KeyCache` on **every continuation page of a storage-map scan**, so each continuation costs an upstream RPC round-trip. On a workload that enumerates storage maps through a fork with a storage overlay, this turns ~20 upstream `state_getKeysPaged` calls into ~2,500 — a **~107× amplification**, and a 31 s read becomes ~400 s.

The regression comes from [#1028](https://github.com/AcalaNetwork/chopsticks/pull/1028) (commit `b1458c4`, released in 1.4.0), specifically the sub-commit *"Refine getKeysPaged guard to bypass cache for off-grid prefixes"*.

**The correctness fix in that commit is right, and I am not asking you to revert it.** The off-grid-prefix hazard is real — I reproduced it on 1.3.1 (see below). But it is justified by the **prefix** check alone; the additional **`startKey`** check is what causes the regression, and it is not needed for correctness.

Ironically, #1028's stated goal (via [polkadot-ecosystem-tests#606](https://github.com/open-web3-stack/polkadot-ecosystem-tests/issues/606)) was to *reduce* upstream round-trips — the old `prefix === startKey` early-return was profiled as costing "one upstream round-trip (~120 ms)" per call.

## The guard

```js
// 1.3.1 — proxy the FIRST page; serve scan CONTINUATIONS from the KeyCache
if (prefix === startKey || prefix.length < minPrefixLen || startKey.length < minPrefixLen) {
  return this.#api.getKeysPaged(prefix, pageSize, startKey, this.#at)
}

// 1.4.0+ — proxy unless BOTH prefix and startKey are EXACTLY minPrefixLen
if (
  prefix.length < minPrefixLen || startKey.length < minPrefixLen ||
  prefix.length > minPrefixLen || startKey.length > minPrefixLen
) {
  return this.#api.getKeysPaged(prefix, pageSize, startKey, this.#at)
}
```

**A scan continuation's `startKey` is a full-length storage key by construction** — it is the last key of the previous page. So `startKey.length > minPrefixLen` is true for every continuation, and the cache path is now reachable **only when `startKey === prefix`, i.e. only for the very first page**.

The net effect is that 1.4.0 populates the in-memory `KeyCache` (and the new on-disk `paged_keys` table) with a map's whole key list, and then never consults either again for the rest of the scan.

## Why it amplifies so hard: the storage overlay

`Block.getKeysPaged` → `block.storage.getKeysPaged(...)`.

- With **no** overlay, `block.storage` **is** the `RemoteStorageLayer`, so a scan is one call with `pageSize` 1000 → one upstream round-trip per 1000 keys.
- With an overlay, `block.storage` is a `StorageLayer`, whose `getKeysPaged` **loops `findNextKey` one key at a time** → `RemoteStorageLayer.findNextKey` → `getKeysPaged(prefix, 1, startKey)`. **One call per key.**

Any config with **`import-storage`** has that overlay. So on 1.4+, enumerating a map through such a fork costs **one upstream round-trip per key**.

**`import-storage` is load-bearing in the repro — without it the regression is invisible** (I initially measured without it and got 32.5 s on 1.3.1 vs 32.4 s on 1.5.0, i.e. no difference at all).

## Minimal repro

`config.yml` (the `import-storage` block is what pushes the overlay — it does not otherwise matter what it sets):

```yaml
endpoint: wss://rpc.hydradx.cloud
block: 12959298
port: 8000
db: ./db.sqlite
import-storage:
  System:
    Account:
      - - - 5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY
        - providers: 1
          data:
            free: '1000000000000000'
```

Then enumerate a single storage map and count what the fork sends upstream:

```js
const api = await ApiPromise.create({ provider: new WsProvider('ws://localhost:8000') })
const entries = await api.query.assetRegistry.assets.entries()
```

Counting `state_getKeysPaged` on the wire between chopsticks and the upstream endpoint:

| chopsticks | keys enumerated | wall clock | **upstream `state_getKeysPaged`** |
|---|---|---|---|
| **1.3.1** | 1,431 | **3.5 s** | **4** |
| **1.5.0** | 1,431 | **342.4 s** | **1,431** |

That is **one upstream round-trip per key** on 1.5.0 (~113 ms each), against 4 paged calls on 1.3.1 — a **~98× slowdown** and **~358× the RPC traffic**, for enumerating a single storage map. (Measured on clean, unmodified installs of each version, counting on the wire between chopsticks and the upstream endpoint.)

The amplification is **linear in the number of keys enumerated** — one upstream round-trip per key.

## Larger workload (same fork, same block, warm `db:`)

A bulk read that enumerates ~10 storage maps (~2,470 keys total):

| chopsticks | `RemoteStorageLayer.getKeysPaged` calls | served from KeyCache | **upstream `state_getKeysPaged`** | wall clock |
|---|---|---|---|---|
| 1.3.1 | 2,469 | 2,460 | **23** | **31 s** |
| 1.5.0 | 2,469 | 9 | **2,468** | **~400 s** |

Counted two independent ways (instrumented inside `RemoteStorageLayer`, and on the wire with a counting WS proxy). It is **not** compute: the fork idles at ~5 % CPU and does the same ~35 s of CPU work on both engines — the extra ~370 s is serialized WAN round-trips at ~113 ms each.

## Affected versions

| version | guard | affected |
|---|---|---|
| 1.3.1 and earlier | serves continuations from the KeyCache | no |
| 1.4.0, 1.4.1, 1.4.2, 1.5.0 | bypasses the cache on any off-grid `startKey` | **yes** |
| `master` (as of 2026-06-26) | unchanged | **yes** |

`storage-layer.ts` and `key-cache.ts` are byte-identical from 1.4.0 through `master`.

## Suggested fix

Keep the correctness bypass for off-grid **prefixes**; drop it for off-grid **startKeys**:

```diff
-    if (
-      prefix.length < minPrefixLen ||
-      startKey.length < minPrefixLen ||
-      prefix.length > minPrefixLen ||
-      startKey.length > minPrefixLen
-    ) {
+    if (prefix.length !== minPrefixLen || startKey.length < minPrefixLen) {
       return this.#api.getKeysPaged(prefix, pageSize, startKey, this.#at)
     }
```

This is safe in both directions: an off-grid prefix is still proxied (so the cache can never *answer* one), and `feed()` is only reachable from the cache path (so an off-grid batch can never *poison* the 66-char group either).

### On the correctness bug #1028 fixed — it is real, and this keeps it fixed

To be sure I was not asking you to reintroduce a bug, I reproduced it against 1.3.1 on a real chain, using a double map (so an account's partial prefix is 162 chars — off the 66-char grouping width): warm the 66-char group with a contiguous run, then ask for the continuation *after an account's last key*. The correct answer is `[]`.

| engine | keys returned | **out-of-prefix** |
|---|---|---|
| 1.3.1 | 5 | **5** — it leaked the *next account's* keys |
| 1.4.0+ / the patch above | 0 | 0 |

So the guard's `prefix` check is doing real work and must stay. Only the `startKey` clause needs to go.

I have a PR ready with this change plus two unit tests in `packages/core/src/blockchain/get-keys-paged.test.ts`: one asserting a scan continuation issues **no** upstream call when the KeyCache already covers the range (this fails on `master`), and one asserting an off-grid prefix is **never** answered from the cache (this protects #1028's fix).

krzysztof-jelski:
Fixes #1054.

## What

Since #1028 (released in 1.4.0), `RemoteStorageLayer.getKeysPaged` bypasses the `KeyCache` on **every continuation page of a storage-map scan**, so each continuation costs an upstream RPC round-trip.

#1028 refined the guard to bypass the cache for **off-grid prefixes** — prefixes longer than the cache's `minPrefixLen` grouping width, which the cache cannot answer correctly (it matches on the truncated head and can return keys that don't start with the requested prefix, tripping smoldot's `key.starts_with(prefix)` assertion in `clear_prefix` paths).

**That correctness fix is right, and this PR keeps it.** I verified the bug it fixed is real: on 1.3.1, an off-grid scan on a double map returned 5 keys, **all 5 outside the requested prefix** (it leaked the next map entry's keys). The `prefix` check is doing real work.

The problem is that the same guard also bypasses the cache when the **`startKey`** is off-grid:

```js
prefix.length > minPrefixLen || startKey.length > minPrefixLen
```

**A scan continuation's `startKey` is a full-length storage key by construction** — it is the last key of the previous page. So this bypasses the cache on *every continuation of every scan*, and the cache path becomes reachable only when `startKey === prefix`, i.e. only for the very first page. 1.4.0 therefore populates the `KeyCache` (and the new on-disk `paged_keys` table) with a map's whole key list, then never consults either again for the rest of the scan.

This inverts #1028's own goal — polkadot-ecosystem-tests#606 profiled the old `prefix === startKey` early-return precisely because it cost "one upstream round-trip (~120 ms)" per call.

## Why it amplifies so hard

`Block.getKeysPaged` → `block.storage.getKeysPaged(...)`. With a storage overlay (i.e. **any config using `import-storage`**), `block.storage` is a `StorageLayer`, whose `getKeysPaged` **walks `findNextKey` one key at a time** → `RemoteStorageLayer.getKeysPaged(prefix, 1, startKey)`. On 1.4+, that makes **each key its own upstream `state_getKeysPaged`**.

Enumerating a **single storage map** (1,431 keys) through such a fork, counted on the wire between chopsticks and the upstream endpoint, on clean unmodified installs:

| chopsticks | keys | wall clock | upstream `state_getKeysPaged` |
|---|---|---|---|
| **1.3.1** | 1,431 | **3.5 s** | **4** |
| **1.5.0** | 1,431 | **342.4 s** | **1,431** |

One upstream round-trip per key — a ~98× slowdown and ~358× the RPC traffic. Full repro in #1054.

## The fix

Restrict the bypass to off-grid **prefixes**:

```diff
-    if (
-      prefix.length < minPrefixLen ||
-      startKey.length < minPrefixLen ||
-      prefix.length > minPrefixLen ||
-      startKey.length > minPrefixLen
-    ) {
+    if (prefix.length !== minPrefixLen || startKey.length < minPrefixLen) {
       return this.#api.getKeysPaged(prefix, pageSize, startKey, this.#at)
     }
```

Safe in both directions:
- an off-grid **prefix** is still proxied, so the cache can never *answer* one (#1028's fix preserved);
- `feed()` is only reachable from the cache path, so an off-grid batch can never *poison* the grouped range either.

## Tests

Two tests added to `packages/core/src/blockchain/get-keys-paged.test.ts`:

1. **`serves scan continuations from the KeyCache instead of the remote`** — asserts a continuation within an already-cached range issues **no** upstream call. **This fails on `master`** with `expected 4 to be 1` (one first page + three continuations, each proxied), and passes with the fix.
2. **`never answers an off-grid prefix from the KeyCache`** — warms the grouped range, then asks for a continuation past an off-grid prefix's last key; asserts `[]` rather than the next entry's keys. This guards #1028's correctness fix against regression.

All pre-existing tests in the file still pass (10/10 with the fix). `biome check` clean on both files.

Verified end-to-end on the real workload: with this change, 1.5.0 goes from ~400 s / 2,468 upstream calls to **31.0 s / 122 upstream calls** on the read that motivated the report, still authors blocks, and still returns identical state.

> Note: `yarn lint` fails in my environment on `packages/core/src/wasm-executor/index.ts` (`Cannot find module '@acala-network/chopsticks-executor'`) because I did not build the Rust executor. It reproduces identically on pristine `master`, so it is unrelated to this change.

### DIFF
--- a/packages/core/src/blockchain/get-keys-paged.test.ts
+++ b/packages/core/src/blockchain/get-keys-paged.test.ts
@@ -543,3 +543,77 @@ describe('RemoteStorageLayer KeyCache', () => {
     for (const key of got) expect(key.startsWith(accountA)).toBe(true)
   })
 })
+
+// @xlc's (maintainer, #1055) reported correctness hole: an on-grid PREFIX (exactly minPrefixLen=66)
+// combined with a startKey drawn from a DIFFERENT 66-char group. The pre-fix guard passed both of these
+// (prefix.length === 66, startKey.length >= 66), so the request reached the cache — and KeyCache.next()
+// picks its range from the first 66 chars of the *startKey*, ignoring the method's prefix, so it returned a
+// key belonging to the other group instead of []. (The `off-grid prefix` test above never covered this: it
+// uses a prefix LONGER than 66, which the guard already proxies.)
+describe('RemoteStorageLayer KeyCache — cross-prefix startKey (#1055)', () => {
+  const hash = '0x1111111111111111111111111111111111111111111111111111111111111111'
+  // Two DIFFERENT exactly-66-char (on-grid) prefixes.
+  const prefixA = '0x1111111111111111111111111111111111111111111111111111111111111111'
+  const prefixB = '0x2222222222222222222222222222222222222222222222222222222222222222'
+  const keys = [`${prefixA}01`, `${prefixA}02`, `${prefixB}01`, `${prefixB}02`]
+
+  const makeLayer = () => {
+    const getKeysPaged = vi.fn(async (p: string, pageSize: number, startKey: string) =>
+      keys.filter((k) => k.startsWith(p) && k > startKey).slice(0, pageSize),
+    )
+    const api = { getKeysPaged } as unknown as Api
+    return { layer: new RemoteStorageLayer(api, hash, undefined), getKeysPaged }
+  }
+
+  it('does not answer prefixA from a startKey that belongs to prefixB', async () => {
+    const { layer } = makeLayer()
+
+    // Warm prefixB's group.
+    await layer.getKeysPaged(prefixB, 10, prefixB)
+
+    // Ask for prefixA, but hand it a startKey from prefixB's group. No key starting with prefixA is
+    // greater than `${prefixB}01`, so the only correct answer is []. The buggy cache returns `${prefixB}02`.
+    const got = await layer.getKeysPaged(prefixA, 10, `${prefixB}01`)
+    expect(got).toEqual([])
+    for (const key of got) expect(key.startsWith(prefixA)).toBe(true)
+  })
+})
+
+// Quantifies the 1.4.0 regression this PR reverses: a single storage-map scan walked one page at a time
+// must cost a small CONSTANT of upstream round-trips (the map is fetched into the KeyCache in one batch and
+// every continuation is served from cache), NOT ~one round-trip per key.
+describe('RemoteStorageLayer KeyCache — continuation upstream call count (#1055)', () => {
+  const hash = '0x1111111111111111111111111111111111111111111111111111111111111111'
+  const prefix = '0x3333333333333333333333333333333333333333333333333333333333333333'
+  const K = 50
+  const keys = Array.from({ length: K }, (_, i) => `${prefix}${i.toString(16).padStart(4, '0')}`)
+
+  const makeLayer = () => {
+    const getKeysPaged = vi.fn(async (p: string, pageSize: number, startKey: string) =>
+      keys.filter((k) => k.startsWith(p) && k > startKey).slice(0, pageSize),
+    )
+    const api = { getKeysPaged } as unknown as Api
+    return { layer: new RemoteStorageLayer(api, hash, undefined), getKeysPaged }
+  }
+
+  it(`scans ${K} keys page-by-page (pageSize 1) with a constant number of upstream calls, not ~${K}`, async () => {
+    const { layer, getKeysPaged } = makeLayer()
+
+    // Walk the whole map one key at a time, exactly as findNextKey does.
+    const scanned: string[] = []
+    let startKey = prefix
+    for (;;) {
+      const page = await layer.getKeysPaged(prefix, 1, startKey)
+      if (page.length === 0) break
+      scanned.push(page[0]!)
+      startKey = page[0]!
+    }
+
+    // Correctness: we saw every key, in order.
+    expect(scanned).toEqual(keys)
+
+    // Perf: the K continuations add zero upstream calls; the whole scan costs a small constant independent of
+    // K. The 1.4.0/1.5.0 regression bypasses the cache on every continuation, costing ~K instead.
+    expect(getKeysPaged.mock.calls.length).toBeLessThan(K)
+  })
+})
--- a/packages/core/src/blockchain/get-keys-paged.test.ts
+++ b/packages/core/src/blockchain/get-keys-paged.test.ts
@@ -494,3 +494,52 @@ describe('getKeysPaged', () => {
     expect(evenKeys, 'evenKeys').toEqual(allKeys.filter((x) => x.startsWith(evenPrefix)))
   })
 })
+
+describe('RemoteStorageLayer KeyCache', () => {
+  const hash = '0x1111111111111111111111111111111111111111111111111111111111111111'
+  // A prefix exactly PREFIX_LENGTH (66) chars wide — the width the KeyCache groups on.
+  const prefix = '0x2222222222222222222222222222222222222222222222222222222222222222'
+  // Two "accounts" under it, i.e. off-grid prefixes: longer than the grouping width, as a double-map's
+  // partial key (or smoldot's clear_prefix) produces.
+  const accountA = `${prefix}aaaa`
+  const accountB = `${prefix}bbbb`
+  const keys = [`${accountA}01`, `${accountA}02`, `${accountB}01`, `${accountB}02`]
+
+  const makeLayer = () => {
+    const getKeysPaged = vi.fn(async (p: string, pageSize: number, startKey: string) =>
+      keys.filter((k) => k.startsWith(p) && k > startKey).slice(0, pageSize),
+    )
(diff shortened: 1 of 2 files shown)


xlc:
I found one cache-boundary correctness issue in the new guard.

A long `startKey` is treated as cache-safe based only on its length, even when it belongs to a different requested prefix. `KeyCache.next()` selects its cache range from the first `minPrefixLen` characters of the `startKey`, and the returned key is not checked against the method's `prefix`.

I reproduced this after warming prefix B:

```ts
await layer.getKeysPaged(prefixB, 10, prefixB)
await layer.getKeysPaged(prefixA, 10, `${prefixB}01`)
```

The second call returns `${prefixB}02` instead of `[]`.

The performance fix can be retained while proxying non-continuation inputs:

```ts
if (
  prefix.length !== minPrefixLen ||
  startKey.length < minPrefixLen ||
  !startKey.startsWith(prefix)
) {
  return this.#api.getKeysPaged(prefix, pageSize, startKey, this.#at)
}
```

Please add a regression test for a warmed unrelated prefix. With the additional guard, that regression plus the existing `get-keys-paged` and `key-cache` suites passed locally (11/11 tests).

krzysztof-jelski:
You're right — the guard trusted a full-length `startKey` without checking it belongs to *this* prefix, so a cross-group `startKey` gets answered from the wrong cache group. Fix applied (`!startKey.startsWith(prefix)`), plus a regression for the warmed-unrelated-prefix case — our existing test only covered an off-grid *prefix*, not this. Perf is unchanged: genuine continuations still hit the cache (verified the upstream call count for a page-by-page scan stays constant, not ~1 per key). Thanks for the catch!

xlc (approved):

